PR #23, opened 2024-03-22
feat: implement quadtree

andrewgazelka:
WIP not really much done here yet

So the goal of this is to implement a quadtree in Rust or at least a similar data structure such that we can query players that are close to each other within a certain distance. Also quadtrees might be useful for partitioning, and I am not aware of many quadtree solutions in Rust that are in crates.io so I think it makes sense to make our own unless we can find one.

# Decisions
- quadtree vs linear quadtree


andrewgazelka:
closing for now.

PR #63, opened 2024-04-01
feat: improve packet handling

TestingPlant:
This includes several changes:
- Packets now have metadata. This metadata can be used to prevent a specific player from receiving a packet and to mark the packet as being allowed to be dropped.
- Packets can be dropped. This is useful to stop clients from going more and more into the past. For example, if a client has a 1 MiB/s download speed and the server is sending the client 2 MiB/s of entity movement packets, the client will be more behind as time progresses. To solve this, some movement packets will be dropped.
- Includes functions to get players' download speed, ping, and send window size which will be useful to determine how much data needs to be dropped.
- Removes the `server_port == 25565` check so it's easier to use a proxy to connect to the server.


andrewgazelka:
Thanks for your work. I will review it tomorrow when I wake up. :)

server/src/singleton/encoder.rs
```diff
@@ -1,99 +1,154 @@
...
     where
         P: Packet + Encode,
     {
-        let start_len = self.buf.len();
+        // Reserve MAX_PACKET_SIZE_LEN bytes before the packet to have room to write the packet
+        // size without shifting the packet body. This means that there is some amount of unused
+        // memory, but the amount of unused memory should be negligible.
+        let mut packet_start = self.packet_data.len();
+        self.packet_data
+            .resize(packet_start + MAX_PACKET_SIZE_LEN, 0);
```
andrewgazelka, line 69: maybe rename MAX_PACKET_SIZE_LEN? kinda seems confusing name

server/src/system/broadcast_packets.rs
```diff
@@ -0,0 +1,150 @@
...
+        // TODO: Avoid taking packet_data so that the capacity can be reused
+        let packet_data = Bytes::from(core::mem::take(&mut encoder.packet_data));
+
+        for (player_uuid, pose, player, _) in &player {
+            let player_location = DVec2::new(pose.position.x, pose.position.y);
+
+            // Max bytes that should be sent this tick
+            // TODO: Determine max_bytes using the player's network speed, latency, and current
+            // send window size
+            let max_bytes = 25_000;
```
andrewgazelka, line 55: 25k bytes * 20 per second

= 4 Mbit/s

hmm think this is fair there should be an issue created from the TODO automatically

server/src/system/broadcast_packets.rs
```diff
@@ -0,0 +1,150 @@
...
+                            .send_raw(packet_data.slice(packet.offset..packet.offset + packet.len))
+                            .is_err()
+                        {
+                            return;
+                        }
+
+                        // total_bytes_sent is not increased because it is no longer used
+                    }
+                } else {
+                    rng.shuffle(&mut encoder.droppable_packets);
```
andrewgazelka, line 98: hmmm, maybe this could be done globally? this seems potentially pretty inefficient to do for every player.

server/src/system/broadcast_packets.rs
```diff
@@ -0,0 +1,150 @@
...
+                    for packet in &encoder.droppable_packets {
+                        if packet.exclude_player == Some(player_uuid.0) {
+                            continue;
+                        }
+
+                        // TODO: Determine chance better
+                        // This currently picks packets closest to the front more often than
+                        // packets in the back. To compensate for this, droppable_packets is
+                        // shuffled, but ideally shuffling shouldn't be necessary.
```
andrewgazelka, line 107: I agree shuffling is bad.. ideally broadcasting thing I am working on will improve this

server/src/io.rs
```diff
@@ -212,6 +234,25 @@ impl IoWrite {
 
         Ok(())
     }
+
```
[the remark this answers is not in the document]
andrewgazelka: or change to your own i used GPT for this

server/src/io.rs
```diff
@@ -381,11 +423,32 @@ impl Io {
...
+            let mut past_queued_send = 0;
+            let mut past_instant = Instant::now();
             while let Ok(bytes) = s2c_rx.recv_async().await {
+                let len = bytes.len();
                 if let Err(e) = io_write.send_packet(bytes).await {
                     error!("{e:?}");
                     break;
                 }
+                let elapsed = past_instant.elapsed();
+                if elapsed > Duration::from_secs(1) {
```
andrewgazelka, line 435: clarify why 1 second?

server/src/io.rs
```diff
@@ -76,9 +91,17 @@ pub struct IoRead {
 pub struct WriterComm {
     tx: flume::Sender<bytes::Bytes>,
     enc: PacketEncoder,
+
+    /// Approximate speed that the other side can receive the data that this sends.
+    /// Measured in bytes/second.
+    speed: Arc<AtomicU32>,
```
andrewgazelka, line 97:
```suggestion
    speed_mib: Arc<AtomicU32>,
```

server/src/io.rs
```diff
@@ -76,9 +91,17 @@ pub struct IoRead {
...
 }
 
 impl WriterComm {
+    pub fn speed(&self) -> u32 {
```
andrewgazelka, line 101:
```suggestion
    pub fn speed_mib(&self) -> u32 {
```

or `mib_per_s` potentially? or could also make this function still be `speed` make this return a type 

struct Speed(u32) (or similar)

and have method mib_per_second()

andrewgazelka (changes requested):
just a few changes :)

server/src/io.rs
andrewgazelka: b7c458c

server/src/io.rs
andrewgazelka: b7c458c

server/src/system/broadcast_packets.rs
andrewgazelka: adding todo

server/src/singleton/encoder.rs
andrewgazelka: fe6f119

server/src/io.rs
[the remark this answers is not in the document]
andrewgazelka: dcbf449

andrewgazelka (approved):

PR #87, opened 2024-04-01
refactor: broadcast crate

andrewgazelka:
Thoughts:

I am _KISS_ for now. I was going to do indexing with Z-order or Hilbert curves, but now I am just doing a naïve approach. This might be good enough. For a square query of  $64 \times 64$ chunks, we will get $64$ byte slices that have $64$ chunks each and slice has length $\text{slice-length}(y) = \Sigma_x l(x,y) \cdot$ where $l$ is the length of the data stored at chunk $x,y$.


TestingPlant (approved):
The broadcast code by itself looks good. I'm not sure how this would be used to make entity movement packets localized in the future, but I'm assuming you have a general idea already.

Issue #61, opened 2024-03-31
add YouTube video to README


andrewgazelka:

andrewgazelka:
closing for now. will wait until get a better video.

PR #95, opened 2024-04-01
docs: add intro video (#61)

TestingPlant:
The README only includes an image of the video thumbnail that needs to be clicked, but there doesn't appear to a better way to embed videos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,10 @@
 [![Discord invite link](https://dcbadge.vercel.app/api/server/PBfnDtj5Wb)](https://discord.gg/PBfnDtj5Wb)
 [![wakatime](https://wakatime.com/badge/github/andrewgazelka/hyperion.svg)](https://wakatime.com/badge/github/andrewgazelka/hyperion)
 
+Introduction (click to watch):
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,10 @@
 [![Discord invite link](https://dcbadge.vercel.app/api/server/PBfnDtj5Wb)](https://discord.gg/PBfnDtj5Wb)
 [![wakatime](https://wakatime.com/badge/github/andrewgazelka/hyperion.svg)](https://wakatime.com/badge/github/andrewgazelka/hyperion)
 
+Introduction (click to watch):
+
+[![Intro video](https://img.youtube.com/vi/hRc2oWAXd70/0.jpg)](https://www.youtube.com/watch?v=hRc2oWAXd70)
```
andrewgazelka, line 8: I think I would make it a clickable webm which is like a GIF but better compression. I have done this a few times so I can try to do it if you want :).

andrewgazelka (changes requested):

README.md
TestingPlant: You can add it as a webm

andrewgazelka:
closing for now

PR #96, opened 2024-04-01
feat: add rayon-local

andrewgazelka:
improves thread-local experience in particular adding to a thread-local queue without being in a rayon context

https://github.com/andrewgazelka/hyperion/blob/0abd642817aced672f2e54da5ec7e6150c81f86f/crates/server/src/system/init_entity.rs#L59-L62


crates/server/src/singleton/encoder.rs
```diff
@@ -135,40 +129,45 @@ impl PacketBuffer {
...
     {
         rayon::broadcast(|_| {
-            let mut encoder = ENCODER.take();
+            let local = self.rayon_local.get_rayon_local();
+            let mut encoder = local.take();
             f(&mut encoder);
-            ENCODER.set(encoder);
+            local.set(encoder);
         });
     }
```
TestingPlant, line 172: Could `rayon::broadcast` be replaced with `self.rayon_local.get_all_locals()`?

TestingPlant (changes requested):

crates/server/src/singleton/encoder.rs
andrewgazelka: yes 5374c88

TestingPlant (approved):

PR #100, opened 2024-04-02
feat: add world border

### LINKED ISSUE #89: add world border

andrewgazelka:
Closes #89


crates/server/src/global.rs
```diff
@@ -1,6 +1,23 @@
...
+use evenio::component::Component;
+
+pub struct Shared {
     // realistically, we will never have more than 2^32 = 4,294,967,296 players
     pub player_count: AtomicU32,
 }
+
+#[derive(Component)]
+pub struct Global {
+    pub world_border_diameter: Option<f64>,
```
herin049, line 12: any reason this is `Option<f64>`?

herin049 (approved):

crates/server/src/global.rs
andrewgazelka: because there can be no world border

PR #101, opened 2024-04-02
perf: use spin mutex for GLOBAL_PACKETS

### LINKED ISSUE #94: ping when running locally is unreasonably high
CleanShot 2024-04-01 at 16 36 43@2x

TestingPlant:
Before this, packets that were received could take 2 ticks to be processed because they need to be moved from LOCAL_PACKETS to GLOBAL_PACKETS. To fix this, LOCAL_PACKETS is removed and a spin lock is used for GLOBAL_PACKETS.

This fixes #94 because the increased ping is a symptom of packets taking up to 2 ticks to be handled. With this fix, ping over localhost should just be the wait_duration for each tick (usually 50ms, but varies slightly)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1767,6 +1767,7 @@ dependencies = [
  "sha2",
  "signal-hook",
  "smallvec 2.0.0-alpha.5",
+ "spin",
  "thread-priority",
  "tracing",
  "tracing-flame",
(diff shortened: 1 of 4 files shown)


andrewgazelka:
thoughts on https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html ?

CleanShot 2024-04-01 at 20 24 45@2x

I guess contention is really low so it might be ok ... 🤔

andrewgazelka (approved):
going to merge for now https://github.com/andrewgazelka/hyperion/issues/102

TestingPlant:
> I guess contention is really low so it might be ok ... 🤔

Since the only time there would be possible contention is when taking GLOBAL_PACKETS each tick, I think using a spin mutex would be fine.

PR #103, opened 2024-04-02
feat: remove unwraps

andrewgazelka:
remove unwraps from several locations in code. In the future, potentially we can have something that wraps `anyhow` (or maybe `thiserror` so we can respond to errors differently depending on the context.


TestingPlant:
IMO we should allow expect so that there's better panic messages if they ever get triggered

andrewgazelka:
> IMO we should allow expect so that there's better panic messages if they ever get triggered

Sure let's do that. however, I think we should try to use `expect` to a limited amount as well.

andrewgazelka:
> IMO we should allow expect so that there's better panic messages if they ever get triggered

02834b7

andrewgazelka:
closing for now

Issue #84, opened 2024-04-01
verify "io_uring supported"


andrewgazelka:
I don't really trust it because it says it is even without seccomp:unconfined

TestingPlant:
Fixed by #188

PR #188, opened 2024-04-09
feat: check if monoio is using io uring (#84)

TestingPlant:
I haven't tested this in an environment that blocks io_uring, but in theory, this should work better than the original check.


andrewgazelka:
oo this looks great

TestingPlant:
Seems to work when using `sysctl kernel.io_uring_disabled=1` to disable io uring system wide

andrewgazelka:
ok fixed for macOS

andrewgazelka:
hmmmm weird behavior don't know why `fix?` commit is needed

PR #193, opened 2024-04-09
fix: don't panic when username is less than 10 chars

TestingPlant:


andrewgazelka:
Oh I just made a PR. https://github.com/andrewgazelka/hyperion/pull/187. However, I will choose this one as I think all usernames are ASCII

PR #207, opened 2024-04-10
fix: write_all_vectored

andrewgazelka:


crates/server/src/net.rs
```diff
@@ -205,32 +207,51 @@ impl IoRead {
...
-            fn read_iovec_len(&self) -> usize {
-                self.iovecs.len()
+impl Buf {
+    fn progress(mut self, mut len: usize) -> Option<Self> {
+        loop {
+            let vec = self.iovecs.get_mut(self.idx)?;
+
+            if len >= vec.iov_len {
+                len -= vec.iov_len;
+                // todo: swap remove
```
TestingPlant, line 237: This todo isn't needed anymore

TestingPlant (changes requested):

crates/server/src/net.rs
```diff
@@ -240,39 +261,21 @@ impl IoWrite {
...
-    pub async fn send_data_macos(
-        &mut self,
-        bytes: ArrayVec<bytes::Bytes, MAX_VECTORED_WRITE_BUFS>,
-    ) -> io::Result<()> {
-        for bytes in bytes {
-            let (result, _) = self.write.write_all(bytes).await;
-            result?; // error occurs here
+        while let Some(buf) = buf_on.take() {
+            let (result, buf) = self.write.writev(buf).await;
+            let len_read = result?;
```
TestingPlant, line 272: We should check if this is 0 to detect the end of the stream

TestingPlant: Also, if the error code is Interrupted, we should retry the send instead of erroring

andrewgazelka:
``` 
RETURN VALUES
     Upon successful completion the number of bytes which were written is returned.  Otherwise, a -1 is returned and the global variable errno is set to indicate the error.
```

I think `len_read`, `0` is returned this means that it was not able to write anything (perhaps for some weird reason) and it might make sense to try again

crates/server/src/net.rs
andrewgazelka: c43bd9e

crates/server/src/net.rs
TestingPlant: Nevermind, I was looking at monoio's code which returned a WriteZero error in this case. https://stackoverflow.com/questions/5656628/what-should-i-do-when-writefd-buf-count-returns-0 states that it should never return 0 unless a 0 byte buffer is passed, but I can't find any sources for that.

TestingPlant (approved):

PR #209, opened 2024-04-10
Add rust-toolchain.toml

Eoghanmc22:
Add a rust-toolchain.toml to the root of the repo to enable the nightly tool chain by default

Fixes: #208


andrewgazelka:
Awesome, this looks good to me. In the future, it might make sense to also have pinned nightly Rust version, especially when we are running this in production, just to make sure that there aren't any regressions.

PR #232, opened 2024-04-11
feat: Use relative movement packets

Eoghanmc22:
Use the correct relative movement packets and only send position updated when needed. Substantially decreases bandwidth and packet count.


andrewgazelka:
winnie-the-pooh-happy-dance-gif-by-leon-denise

andrewgazelka:
> Use the correct relative movement packets and only send position updated when needed. Substantially decreases bandwidth and packet count.

image
I don't see `needs_resync` being set to true anywhere. Is this planned to impl in the future?

Eoghanmc22:
Its not really needed rn. Just put it there as an api in case its needed in future.

PR #273, opened 2024-04-14
perf: use io uring

TestingPlant:


andrewgazelka (changes requested):
> for the readme, i think it might actually need linux 6.1

andrewgazelka (approved):
approving for now even though it is buggy

PR #297, opened 2024-04-19
fix: io several uring errors

andrewgazelka:


crates/server/src/net/linux.rs
```diff
@@ -94,9 +99,19 @@ unsafe impl Sync for LinuxServer {}
...
+        let domain = match address {
+            SocketAddr::V4(_) => socket2::Domain::IPV4,
+            SocketAddr::V6(_) => socket2::Domain::IPV6,
+        };
+
+        let listener = Socket::new(domain, socket2::Type::STREAM, None)?;
+        listener.set_nonblocking(true)?;
+        listener.set_send_buffer_size(SEND_BUFFER_SIZE)?;
+        listener.bind(&address.into())?;
+        listener.listen(LISTEN_BACKLOG)?;
```
andrewgazelka, line 114: this is the number of kernel write requests that can be had to a given FD, right?

PR #299, opened 2024-04-21
Allow attacking zombies

Ruben2424:
Hi,
the last few days i played a bit around with the code. 
I made a few changes which allow to attack zombies.

At when the zombies are pushed back from the attack, they don't fall back to the ground.
Do we need a physics engine for that? Is that needed for the event?


andrewgazelka:
Hello Ruben, thank you for your changes. There is no physics engine yet. Part of that is because, while zombies are really good for demonstrations, what we're targeting right away is going to be a player versus player event. We'll still need physics for sure, for instance, for arrows and snowballs if we have those. But at least from what I'm thinking, there won't be a ton of entities, but this is to be determined.

andrewgazelka:
I will look at this once https://github.com/andrewgazelka/hyperion/pull/297/files is merged

andrewgazelka:
LGTM :) great first PR. Just merged it in with `main`.

PR #322, opened 2024-04-24
feat: zstd

andrewgazelka:


andrewgazelka:
libdeflate is good enough probably... and doesn't require a custom client (so more tooling as well) 👍

PR #382, opened 2024-04-30
fix: handle truncated writes

TestingPlant:
This draft currently supports resending truncated packets if only one buffer is sent per player per tick (in theory; haven't tested this). However it doesn't handle the event where the server does the following:
```
write(100 bytes)
write(10 bytes)
submit()
```
and the completion events have:
```
write(100 bytes) = 50 bytes written
write(10 bytes) = 10 bytes written
```
I'm not sure what the best way to solve this would be.

### DIFF
--- a/crates/server/src/net.rs
+++ b/crates/server/src/net.rs
@@ -1,6 +1,6 @@
 //! All the networking related code.
 
-use std::{cell::RefCell, hash::Hash, net::SocketAddr};
+use std::{cell::RefCell, collections::VecDeque, hash::Hash, net::ToSocketAddrs, net::SocketAddr, sync::Mutex};
 
 use anyhow::Context;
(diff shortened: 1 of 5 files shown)


TestingPlant:
To fix the edge case mentioned above, I'm thinking of merging all the packets into one buffer per player and only sending that buffer every tick. This would also be necessary if we decided to add encryption in the server directly without a reverse proxy.

andrewgazelka:
ok this seems great

andrewgazelka:
> To fix the edge case mentioned above, I'm thinking of merging all the packets into one buffer per player and only sending that buffer every tick. This would also be necessary if we decided to add encryption in the server directly without a reverse proxy.

https://github.com/andrewgazelka/hyperion/tree/feat-iou-multiple-threads

PR #386, opened 2024-05-01
feat: trackable player inventory

Ruben2424:


Ruben2424:
The Players now have a `PlayerInventory` component to store the items. 
The current status is usable. At the moment it is possible:
- to give and remove items within the source code
- players see which item other players have in the hand
- to give players items with `/give`

I think the biggest todo before this PR can be merged is to support the moving/splitting of items in the inventory. And to other inventories like chests or something.

andrewgazelka:
looks good so far :)

Ruben2424:
It mostly works now.
Players can move the items in the inventory.
The eqipment like armor is shown to other players.

At the moment The `ClickMode` and `button` does not really have an effect. 
It just applies the slot changes sent by the client and checks if the sum of the items is still correct. I am not sure if this is the best way to do it.

Open Todos are:
- [ ] make sure that there are no edge cases which allow clients to create new items
- [ ] unittests for inventory
- [ ] cleanup

andrewgazelka:
great work :)

I am ready to merge when we are decently confident.

>  make sure that there are no edge cases which allow clients to create new items

I could see this being a big issue in an event.

andrewgazelka:
I improved a couple things; merging :)

PR #429, opened 2024-05-18
perf: use bumpalo for BVH

andrewgazelka:
CleanShot 2024-05-18 at 17 25 03@2x

CleanShot 2024-05-18 at 17 58 47@2x

Looks to be small% faster or so compared to https://youtu.be/2dGtW7E0B3U?si=BziIclI4EfJi81SC


andrewgazelka:
closing this for now. might re-impl in future

PR #433, opened 2024-05-19
feat: arrows

andrewgazelka:
A lot of work still needs to be done on this. Right now, there's basically no code, and physics has to be implemented first. If someone wants to continue this, that would be amazing.


andrewgazelka:
thanks so much for helping work on this @TestingPlant

andrewgazelka:
Really great work, is there any reason we can't merge this in? I'd be fine merging this in now and then having another PR to try to make them more like vanilla because I definitely undershot right now.

TestingPlant:
I think this would be fine to merge in now to let others work on code that need entity physics like for dropped items. There's a few unresolved issues though: 
- arrows don't collide with entities
- bow being drawn back doesn't increase speed
- the code for entity collisions on blocks assumes that all entities are a small point instead of a box
- the server figures out the correct block collision point but the client still sees the arrow falling through blocks when shooting straight up and letting the arrow fall down

andrewgazelka:
ok made issues for all of these

PR #437, opened 2024-05-25
feat: Able to drop items

Ruben2424:
This is a WIP PR to support dropping and picking up items.
At the moment it is mostly a working prototype and still a lot of work todo.

For it to completely work there is also probably the physics #433  necessary so items can fall to the 

- [x] Drop right item(s) when pressing Q/strg Q
- [ ] Send entity packets to all players and to new joining players
- [ ] make sure a item can only picked up once
- [ ] stack items properly 
- [ ] Do not pickup items when inventory is full
- [ ] Drop items from inventory
- [ ] Figure out how to make the `generic_collision` system best
- [ ] Drop items on death?
- [ ] physics and collision with blocks


andrewgazelka:
repo has diverged significantly

Issue #447, opened 2024-05-30
bow being drawn back doesn't increase speed


andrewgazelka:

eduxstad:
close?

andrewgazelka:
closing

PR #453, opened 2024-06-01
feat: bow draw

eduxstad:
Addresses #447 

https://github.com/andrewgazelka/hyperion/assets/15882635/197d8d9e-8ca7-42fb-9925-9fc88fc1d421

### DIFF
--- a/crates/server/src/inventory.rs
+++ b/crates/server/src/inventory.rs
@@ -1,4 +1,4 @@
-use std::{mem, ops::RangeInclusive};
+use std::{mem, ops::RangeInclusive, time::SystemTime, time::UNIX_EPOCH, time::Duration, time::SystemTimeError};
 
 use anyhow::ensure;
 use evenio::component::Component;
@@ -60,6 +60,8 @@ impl<const T: usize> Inventory<T> {
     pub const fn get_items(&self) -> &[ItemStack; T] {
         &self.slots
     }
+
+    
 }
 
 /// The player's inventory.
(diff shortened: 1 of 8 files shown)


crates/server/src/inventory.rs
```diff
@@ -428,6 +432,14 @@ impl PlayerInventory {
                 | ItemKind::Air
         )
     }
+
+    pub fn interact(&mut self) {
+        self.interact_time = SystemTime::now();
```
eduxstad, line 437: Do we want to use SystemTime or do we want to use Gameticks?

I think gameticks will be more performant but may be less consistent.

crates/server/src/inventory.rs
```diff
@@ -73,6 +75,7 @@ pub struct PlayerInventory {
     ///
     /// This item will be none when player closes inventory
     carried_item: ItemStack,
+    interact_time: SystemTime,
```
eduxstad, line 78: Put this in inventory since it can apply to multiple items.

crates/server/src/inventory.rs
TestingPlant: I'd suggest using gameticks since other code already uses them for timing

andrewgazelka: Well, let's consider what would happen if the game ticks in the server were really laggy. People would charge their bow to full, and then the server would think that the bow isn't full. This is bad. We should try to use system time, which should be more accurate, and then we can go to a game tick-based system if there is a performance issue, but I don't think there will be.

PR #457, opened 2024-06-02
fix: use correct arrow offset

TestingPlant:


andrewgazelka:
I don't know if this is related to this commit, but it seems like arrows sink into the floor.

image

andrewgazelka:
Also, it appears to me that when I'm shooting up at things, like at a bridge, sometimes the arrows can phase through your block if it's on the edge.

andrewgazelka:
Going to merge this in now. I just think we need better collision code, and this will be good as a separate commit/PR.

PR #532, opened 2024-10-29
fix: automatic map download

CuzImClicks:
reverts the rename so everything works fine again


andrewgazelka:
merging this in but looking to fix this in the future (and not have it named NewYork)

PR #535, opened 2024-10-30
fix: adding to empty slot and max_stack_size

CuzImClicks:
Fixes try_add_to_slot creating stacks above the maximum stack size when adding to an empty slot


CuzImClicks:
also while working on the give command i noticed that the `try_add_item` method doesn't work correctly because it also fills the 2x2 crafting and armor slots. is there a reason why `try_add_to_slot` is private btw?

andrewgazelka:
>  is there a reason why try_add_to_slot is private btw?

not really

PR #612, opened 2024-11-28
fix(simd-util): allow different align offsets 

### LINKED ISSUE #608: potential SIMD alignment crash?
If the server crashes please give log/stacktrace

> Did you change something about xp handling which broke hyperion?
> Huh
> Wtf it works now
> sus

TestingPlant:
This handles the case where `prev` and `current` from `copy_and_get_diff` have
different align offsets. On my machine (aarch64-unknown-linux-gnu) the code
still usually takes the "optimized" path where the align offsets are the same
(`before_prev.len() == before_current.len()`), so this shouldn't cause significant
performance issues.

Fixes #608


andrewgazelka:
> This handles the case where `prev` and `current` from `copy_and_get_diff` have different align offsets. On my machine (aarch64-unknown-linux-gnu) the code still usually takes the "optimized" path where the align offsets are the same (`before_prev.len() == before_current.len()`), so this shouldn't cause significant performance issues.
> 
> Fixes #608

kinda confused why the offsets would ever be different though... I definitely had it happen a couple times on `aarch64-apple-darwin` feel we should fix this as well 🤔

TestingPlant:
> kinda confused why the offsets would ever be different though... I definitely had it happen a couple times on `aarch64-apple-darwin` feel we should fix this as well 🤔

In sync_entity_state.rs, `prev_xp` and `xp` are `&[u16]`s, meaning that they are aligned to 2 bytes. Let's also assume that `Simd<u16, 64>` has an alignment of 8 bytes (made up number for this example). If `prev_xp` and `xp` are not already aligned to 8 bytes by chance, `as_simd` (which is just `align_to` internally) will need to skip some elements until it's aligned with `Simd`. The number of elements that need to be skipped can vary. For example, if `prev_xp` has address `0x2`, 3 `u16`s need to be skipped until we could use an aligned read to get a `Simd` from address `0x8`. If `prev_xp` had an address of `0x6` then it would only need to skip one `u16` before being able to do an aligned read of a `Simd`. The amount of elements that need to be skipped to perform an aligned read is the align offset, and since `&[u16]` only needs to be aligned to 2 bytes, the align offset to get 8 byte alignment can vary.

andrewgazelka:
> > kinda confused why the offsets would ever be different though... I definitely had it happen a couple times on `aarch64-apple-darwin` feel we should fix this as well 🤔
> 
> In sync_entity_state.rs, `prev_xp` and `xp` are `&[u16]`s, meaning that they are aligned to 2 bytes. Let's also assume that `Simd<u16, 64>` has an alignment of 8 bytes (made up number for this example). If `prev_xp` and `xp` are not already aligned to 8 bytes by chance, `as_simd` (which is just `align_to` internally) will need to skip some elements until it's aligned with `Simd`. The number of elements that need to be skipped can vary. For example, if `prev_xp` has address `0x2`, 3 `u16`s need to be skipped until we could use an aligned read to get a `Simd` from address `0x8`. If `prev_xp` had an address of `0x6` then it would only need to skip one `u16` before being able to do an aligned read of a `Simd`. The amount of elements that need to be skipped to perform an aligned read is the align offset, and since `&[u16]` only needs to be aligned to 2 bytes, the align offset to get 8 byte alignment can vary.

wait ahhhh I think I know the issue... should have been `32` lanes instead of `64` ... we are aligning to 64 bytes

https://github.com/andrewgazelka/hyperion/blob/6ecfd8daa665ec6f007e9a124436395a97fe7f19/crates/hyperion/src/alloc.rs#L11

andrewgazelka:
yea getting

> current slice is not 64-byte aligned even though flecs allocator should be 64-byte aligned

when I add check. am I not understanding someting?

andrewgazelka:
maybe flecs reuses allocations or something 🤔

andrewgazelka:
> > maybe flecs packs multiple things into its allocations 🤔
> 
> I added some prints to `aligned_malloc` and it seems like `xp` and `prev_xp` are from the same allocation. I don't think there's a good way to get flecs to get 64 byte alignment on `xp` and `prev_xp` so handling different align offsets is likely necessary.

andrewgazelka:
> > maybe flecs packs multiple things into its allocations 🤔
> 
> I added some prints to `aligned_malloc` and it seems like `xp` and `prev_xp` are from the same allocation. I don't think there's a good way to get flecs to get 64 byte alignment on `xp` and `prev_xp` so handling different align offsets is likely necessary.

image

andrewgazelka:
ok @TestingPlant I think your approach is fine since doesn't seem to be a good way to solve this.

PR #615, opened 2024-11-29
Fix inventory clear bug when selecting class

Kumpelinus:
Fixes one of the issues of https://github.com/andrewgazelka/hyperion/issues/613

### DIFF
--- a/events/tag/src/command/class.rs
+++ b/events/tag/src/command/class.rs
@@ -48,6 +48,18 @@ impl MinecraftCommand for ClassCommand {
                 &mut Team,
                 &mut Class,
             )>(|(stream, uuid, position, yaw, pitch, team, class)| {
+                if *team == team_param && *class == class_param {
+                    let chat_pkt = agnostic::chat("§cYou’re already using this class!");
+
+                    let mut bundle = DataBundle::new(compose);
+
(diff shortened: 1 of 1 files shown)


events/tag/src/command/class.rs
```diff
@@ -48,6 +48,18 @@ impl MinecraftCommand for ClassCommand {
                 &mut Team,
                 &mut Class,
             )>(|(stream, uuid, position, yaw, pitch, team, class)| {
+                if *team == team_param && *class == class_param {
+                    let chat_pkt = agnostic::chat("§cYou’re already using this class!");
+
+                    let mut bundle = DataBundle::new(compose);
```
Kumpelinus, line 54: Idk if you want to use the DataBundle here or directly send the packet. What is your preferred way of doing this @andrewgazelka?
*Another alternative would be to create the DataBundle at the top and use it for both cases*

andrewgazelka: @Kumpelinus data bundle is for combining multiple packets usually

so you can also just do `compose.unicast` directly like

https://github.com/andrewgazelka/hyperion/blob/1ad208c22f64dd429157543c09b749ad89ae7746/crates/hyperion/src/egress/sync_entity_state.rs#L261

but it doesn't really matter as long as it works

PR #620, opened 2024-12-02
perf: remove regex

Tebarem:
RegEx is really slow. If used in something like chat commands where they are potentially spammed all the time it can lead to bad performance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,7 +1794,6 @@ dependencies = [
  "gxhash",
  "hyperion",
  "indexmap",
- "regex",
  "tracing",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -190,10 +190,6 @@ version = '0.16.1'
 default-features = false
 version = '0.3.6'
 
-[workspace.dependencies.regex]
-features = ['perf-dfa-full']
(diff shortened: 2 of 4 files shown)


andrewgazelka:
Looks good to me :) 

thank you for your work. Btw I am off Discord until Friday (in order to be productive) so if you have any questions I will respond faster on GitHub (Issue/PR/Discussions).

andrewgazelka:
> All good mate! We had a argument on discord i think u should take a look at it maybe give us your thoughts on it. But basically my idea is that instead of using LMDB for permissions and user groups we just have a sqlx client handling it and that enables people to use sqlite or any other sql db while still being lighting fast. This would also let us easily create a global permissions registry and commands registry

ok also in future:

I would prio getting the event done rather than mico optimizations. While this is useful and I appreciate it, there are a lot of other things that beyond being slightly bad performance will make it so we can't even have the event. I wanna get the event and then optimize as needed based on benchmarks :)

Tebarem:
100% i wouldnt consider this to be micro optimizations really. SQLx would be slower than LMDB but it would open up much more things you would be able to do with permissions and commands. The regex thing was a micro optimization i would say

Issue #613, opened 2024-11-28
Fix issues caused by class selection


Kumpelinus:
If you try to "switch" to the class that you already have (of your current team) it does not give you the kit.
Instead it completely clears your inventory.

Kumpelinus:
It seems to be just a wrong packet being sent (or not being sent) since other players can see you holding the items while you can't see them yourself.

Kumpelinus:
I found a few more Issues with this as well.
* It also resets the xp in the client's xp bar. However it does not reset the xp stat on the server so you will get it back upon gaining xp again (So it is just visual). 
* It removes you from your own tab-list which might indicate further problems (can't confirm that yet).

andrewgazelka:
> I found a few more Issues with this as well.
> 
> * It also resets the xp in the client's xp bar. However it does not reset the xp stat on the server so you will get it back upon gaining xp again (So it is just visual).
> * It removes you from your own tab-list which might indicate further problems (can't confirm that yet).

yea changing the lpayer skin requires "killing and respawning htem" client side which can break al ot of thing... we have to send all the packets including XP packets to make sure their state after respawn is the same as before respawn

Kumpelinus:
The class selection itself resets your inventory.
And the code

```rust
.observer::<flecs::OnSet, (
    &Xp,                  //                  (0)
    &UpgradedTo,          //                  (1)
    &Class,               //                  (2)
    &Team,                //                  (3)
    &MainBlockCount,      //                  (4)
    &mut PlayerInventory, //             (5)
)>()
```
normally observes this and thus gives you the equipment you the equipment.

But since neither your team nor your class changes in the case of selecting the same class + team it does not trigger the `apply_inventory` method and that leads to your inventory staying empty.

We could either:
* change the observer to react to an event by the command **OR**
* just check for the same team + class when running the command.

I prefer the latter one since you can also add a error message very easily that way.

andrewgazelka:
pretty sure this issue is fixed now

PR #631, opened 2024-12-03
Re-sync xp bar on changing skin

Kumpelinus:
Relates to #613

### DIFF
--- a/events/tag/src/command/class.rs
+++ b/events/tag/src/command/class.rs
@@ -8,15 +8,14 @@ use hyperion::{
         player_join::{PlayerListActions, PlayerListEntry, PlayerListS2c},
     },
     net::{Compose, DataBundle, NetworkStreamRef, agnostic},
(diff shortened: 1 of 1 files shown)


andrewgazelka:
lgtm ... definitely want to profile if updating inventory often will become a performance issue with 10k players at ome point or not.

Kumpelinus:
Probably adding a cooldown + making a player unable to switch class while in combat might make sense.

andrewgazelka:
> Probably adding a cooldown + making a player unable to switch class while in combat might make sense.

players will only be able to switch classes once they get a number of points to be able to switch... so they won't be able to more than once an average every 5 minutes or so

(if you play diep.io / deeep io / mope io it is basically the same way)

I don't think we should add a cooldown for now

Kumpelinus:
Oh, I thought they would maybe be able to switch back to a previous class. If you can't, it does not make sense to add it.

PR #634, opened 2024-12-04
feat: GUI with click handling and custom item slots

Tebarem:
PLEASE review carefully as i have changed things that MIGHT be important


andrewgazelka:
ok looking @ this

PR #641, opened 2024-12-04
feat: add `ClickMode` to GUI API

Tebarem:
ClickMode for flexibility on how the player clicked an item on the GUI

Also there is a bug where the player is able to shift click and take the items. Not sure how to prevent this.

### DIFF
--- a/crates/hyperion-gui/src/lib.rs
+++ b/crates/hyperion-gui/src/lib.rs
@@ -164,7 +164,7 @@ impl Gui {
                     return;
                 }
 
-                let slot = event.slot_idx as usize;
+                let slot = usize::from(event.slot_idx);
                 let Some(item) = items.get(&slot) else {
                     return;
                 };
--- a/crates/hyperion-gui/src/lib.rs
+++ b/crates/hyperion-gui/src/lib.rs
@@ -10,6 +10,7 @@ use hyperion::{
     valence_protocol::{
         ItemStack, VarInt,
         packets::play::{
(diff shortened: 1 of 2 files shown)


Kumpelinus:
Uhh. This contains quite a few commits from the actual main branch (I guess since you merged instead of rebasing). I am not sure if that may be a problem since it gets squashed anyways but maybe it should be rebased properly to make the PR be less cluttered.

crates/hyperion-gui/src/lib.rs
```diff
@@ -157,17 +158,21 @@ impl Gui {
...
             let gui = self.clone();
             event_handlers.click.register(move |query, event| {
+                let button = event.mode;
+
                 if event.window_id != window_id {
                     return;
                 }
 
-                let slot = usize::from(event.slot_idx);
+                let slot = event.slot_idx as usize;
```
Kumpelinus, line 167: Wasn't From::from the preferred way since it checks for compatible types on compile time instead of just doing it? (Probably does not matter here explicitly since they are probably compatible)

Kumpelinus:
Looks good to me. But you gotta wait for @andrewgazelka.

crates/hyperion-gui/src/lib.rs
andrewgazelka: yea this is correct. I am changing it back. although curious that the lint does not seem to be active anymore. oh well.

andrewgazelka (approved):

PR #642, opened 2024-12-04
feat(gui): sync inventory state after clicks

Tebarem:
Automatically refresh player inventory display after GUI interactions to maintain consistency between GUI and inventory states.

Changes:
- Implement inventory refresh logic using `InventoryS2c` packet


Kumpelinus:
Depends on #641

PR #647, opened 2024-12-05
fix: raycast traversal

Tebarem:
`Ray::voxel_traversal` and `get_direction_from_rotation` were implemented wrong.


andrewgazelka (approved):
going to guess this is right

andrewgazelka:
thanks :)

PR #666, opened 2024-12-08
feat: bows 🏹

Tebarem:
Please someone fix Velocity, Yaw, and Pitch. It was working before systems so im not sure what could've broken it.


andrewgazelka:
Yay will look at this later today

crates/hyperion/src/egress/sync_entity_state.rs
```diff
@@ -237,7 +237,7 @@ impl Module for EntityStateSyncModule {
                         entity_id,
                         #[allow(clippy::cast_possible_truncation)]
                         delta: (position_delta * 4096.0).to_array().map(|x| x as i16),
-                        on_ground: false,
+                        on_ground: velocity.velocity == Vec3::ZERO,
```
andrewgazelka, line 240: Ok probably works for it but would like this to be more accurate in future. Perhaps we should make an issue for it.

crates/hyperion/src/simulation/bow.rs
```diff
@@ -0,0 +1,34 @@
+use std::time::{Duration, SystemTime};
+
+use flecs_ecs::prelude::*;
+
+#[derive(Component, Debug)]
+pub struct BowCharging {
+    pub start_time: SystemTime,
```
andrewgazelka, line 7: Good use of system time probably. Let's make a low priority issue to test with artificial inconsistent latency to see how inaccurate or accurate bows are. I wonder if there's a way to see with TCP at what time bytes were actually sent. I want to do more testing with artificial latency in general but should be low priority

crates/hyperion/src/simulation/handlers.rs
```diff
@@ -328,10 +337,20 @@ pub fn player_interact_item(
...
+            let packet = play::OpenWrittenBookS2c { hand };
+            query.compose.unicast(&packet, query.io_ref, query.system)?;
+        } else if cursor.item == ItemKind::Bow {
+            // Start charging bow
+            let entity = query.world.entity_from_id(query.id);
+            entity.get::<Option<&BowCharging>>(|charging| {
+                if charging.is_some() {
+                    return;
+                }
+                entity.set(BowCharging::new());
```
andrewgazelka, line 351: Generally setting and removing components frequently is not ideal.

andrewgazelka (commented):
Some things I wanna look into

crates/hyperion/src/simulation/bow.rs
andrewgazelka: added issue

PR #670, opened 2024-12-09
feat: vanish 💨

Tebarem:
added `/vanish` command for admins.


Kumpelinus:
@Tebarem did you add the bow charging functionality in the same commit btw? Shouldn't that be 2 different commits / PRs?

Kumpelinus:
Apart from that it looks pretty good to me. I will am working on generalizing the state sync in a different PR btw.

andrewgazelka:
> Apart from that it looks pretty good to me. I will am working on generalizing the state sync in a different PR btw.

generalizing the state sync? some of this is already done (look at 5f641fc) but yea generalizing it even more would be cool. just making sure you are aware :)

PR #695, opened 2024-12-10
feat(core): send more combat related packets

GroobleDierne:
Changes:
- Send HealthUpdateS2c to update health on the client
- Send DamageTiltS2c to trigger the damage tilt on the client (still needs to calculate where the damage comes from)
- Send EntityDamageS2c to display a red outline hover the client taking damage
- Send knockback sound to nearby clients
- Send DeathMessageS2c to trigger the client's respawn sequence
- Broadcast EntityStatusS2c when a player is killed
- Set killed player pose to `Dying`

Missing:
- Handle respawn
- Handle critical hits

### DIFF
--- a/events/tag/src/module/attack.rs
+++ b/events/tag/src/module/attack.rs
@@ -8,13 +8,25 @@ use flecs_ecs::{
 };
 use hyperion::{
     net::{
-        agnostic, packets::{BossBarAction, BossBarS2c}, Compose, ConnectionId
+        Compose, ConnectionId, agnostic,
+        packets::{BossBarAction, BossBarS2c},
     },
-    simulation::{event, metadata::{entity::Pose, living_entity::Health}, PacketState, Player, Position, Velocity},
+    simulation::{
+        PacketState, Player, Position, Velocity, event,
+        metadata::{entity::Pose, living_entity::Health},
+    },
     storage::EventQueue,
     uuid::Uuid,
     valence_protocol::{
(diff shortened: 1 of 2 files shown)


events/tag/src/module/attack.rs
```diff
@@ -164,17 +166,51 @@ impl Module for AttackModule {
...
+                                    };
+                                    let pkt_hurt = play::DamageTiltS2c {
+                                        entity_id: VarInt(target.minecraft_id()),
+                                        yaw: 0. // Todo look at how this is calculated
+                                    };
+                                    // EntityDamageS2c: display red outline when taking damage (play arrow hit sound?)
+                                    let pkt_damage_event = play::EntityDamageS2c {
+                                        entity_id: VarInt(target.minecraft_id()),
+                                        source_cause_id: VarInt(origin.minecraft_id() + 1), // this is an OptVarint
+                                        source_direct_id: VarInt(origin.minecraft_id() + 1), // if hit by a projectile, it should be the projectile's entity id
```
andrewgazelka, line 183: do we have to add 1 because if it is 0 that means there is no source?

andrewgazelka (approved):

events/tag/src/module/attack.rs
GroobleDierne: Exactly, it's just a weird protocol thing, they didn't want to use a boolean prefixed varint like they do most of the time for some reason.

PR #702, opened 2024-12-12
feat: respawn

GroobleDierne:
Current state:
- Works with and without enable_respawn_screen
- Restores player's health
- Change the player's team

Todo
- Respawn next to a teammate

When respawned, the client tries to teleport to its respawn location as defined by PlayerSpawnPositionS2c, but the server prevents this as it exceeds the maximum distance that the client is allowed to move in one tick. This doesn't matter for the Tag event as we'll have to teleport the player to another location anyway, but it could be an issue for other events, so we should have a way to bypass this distance check when it follows a respawn.


andrewgazelka:
> Current state:
> 
> * Works with and without enable_respawn_screen
> * Restores player's health
> * Change the player's team
> 
> Todo
> 
> * Respawn next to a teammate
> 
> When respawned, the client tries to teleport to its respawn location as defined by PlayerSpawnPositionS2c, but the server prevents this as it exceeds the maximum distance that the client is allowed to move in one tick. This doesn't matter for the Tag event as we'll have to teleport the player to another location anyway, but it could be an issue for other events, so we should have a way to bypass this distance check when it follows a respawn.

I don't think we should bypass the distance check; instead we should set the player location properly before we send the respawn packet. bypassing the distance check is a big vulnerability. For instance:

- player could decide they just wanna teleport to any place in the map
- player could just not respawn at all and continue with respawn gear but in same place as they died

andrewgazelka:
@SanderMertens thoughts on 

https://github.com/andrewgazelka/hyperion/blob/5cdd9e36b85369f1e77f67277f32bc65b28c2a69/crates/respawn/src/lib.rs#L34-L35

vs not having `Pose` in the query and doing `.set`?

GroobleDierne:
Just rebased everything.

@andrewgazelka what is the preferred way to get every player on a specific team?
Also `client.modified::<Pose>();` doesn't work properly, the client switch to the dying pose correctly but won't go back to standing.

andrewgazelka:
> @andrewgazelka what is the preferred way to get every player on a specific team? 

use `.query` (similar to system DSL)

could do
```rust
    world
        .query::<&Team>()
        .each_entity(|_, team|) { 
             if team != target continue 
             // ...
         }
```

inside of a system.  see https://github.com/Indra-db/Flecs-Rust/tree/main/flecs_ecs/examples/flecs/queries

>Also `client.modified::<Pose>();` doesn't work properly, the client switch to the dying pose correctly but won't go back to standing.

just do what you were doing then

andrewgazelka:
https://github.com/user-attachments/assets/c4e8fdab-9c22-443d-aa42-e4113acd626f

still appears a bit bugged :(

also—"Respawn next to a teammate" can probably be done in a future PR. let's try to get this in.

GroobleDierne:
> Screen.Recording.2024-12-18.at.9.34.50.AM.mov
> 
> still appears a bit bugged :(
> 
> also—"Respawn next to a teammate" can probably be done in a future PR. let's try to get this in.

This was introduced by [a5c2f89](https://github.com/hyperion-mc/hyperion/pull/702/commits/a5c2f89fa3a6d30313251c38fa25d5d89bc903c6) I'll look at what caused that.
Also there is a new very weird regression with this commit: 

https://github.com/user-attachments/assets/b643b845-768c-485f-84f8-d6d5449c59c0

andrewgazelka:
> > Screen.Recording.2024-12-18.at.9.34.50.AM.mov
> > still appears a bit bugged :(
> > also—"Respawn next to a teammate" can probably be done in a future PR. let's try to get this in.
> 
> This was introduced by [a5c2f89](https://github.com/hyperion-mc/hyperion/pull/702/commits/a5c2f89fa3a6d30313251c38fa25d5d89bc903c6) I'll look at what caused that. Also there is a new very weird regression with this commit:
> 
>  hotbar.bug.mp4

hmmmmm yea in the future I want to invest more in testing infrastructure I think... like having a bot that can connect to the server and we can control it with functions.

GroobleDierne:
Seems correct now

PR #720, opened 2024-12-15
feat(core): Entity To Entity Raycasting.

Tebarem:


Tebarem:
@andrewgazelka if you can update ti a bit to actually use bvh cause i couldn't figure it out in ur impl

crates/hyperion/src/egress/sync_entity_state.rs
```diff
@@ -442,7 +440,6 @@ impl Module for EntityStateSyncModule {
                         velocity.0 = velocity.0.clamp_length(0.0, 100.0);
```
andrewgazelka, line 440: no magic values (define in a constant)

andrewgazelka (approved):

PR #792, opened 2024-12-27
feat: color chat based on team

GroobleDierne:


andrewgazelka:
im confused what is up with benchmarks here lol but looks good to me

PR #795, opened 2024-12-29
fix: don't change skin

GroobleDierne:
Don't change skin based on class as discussed [here](https://discord.com/channels/1222556172385783890/1318666942625747054) 

Also removed https://github.com/hyperion-mc/hyperion/blob/main/events/tag/src/skin.rs


andrewgazelka (approved):
lgtm. I think removing `skin.rs` is a good option as it is not useful anymore. We can always look back in history and re-add it back or add it as a separate crate.

PR #809, opened 2025-01-13
ci: codecov token not needed

andrewgazelka:


andrewgazelka:
In general, I'm really confused right now because Codeco says I don't need a token. However, it is requiring a token it appears

iTerm2 2025-01-13 13 12 31

PR #810, opened 2025-01-14
fix: rust-mc-bot for windows

CuzImClicks:
im gonna test this when i get home


CuzImClicks:
tested it at home and it works now

PR #813, opened 2025-01-14
ci(bench): work on forks

andrewgazelka:


andrewgazelka:
didn’t make a comment unsure it actually works

PR #827, opened 2025-01-21
refactor: restructure events

TestingPlant:
Definitions to avoid ambiguity:
- "Events" - Events from `crates/hyperion/src/simulation/event.rs` that are currently processed through `EventQueue`
- "Event code" - Code in `events/`, such as the `tag` event

Events are mostly repackaged Valence packets with some processing, which can easily be done in the event code. Custom events can also already be sent through `HandlerRegistry`, and keeping the separate events system adds confusion on which system to use. In addition, the `EventQueue` approach requires allocations and requires us to maintain `RuntimeLifetime`, which cannot be checked at compile time and causes aborts at runtime if used improperly while the `HandlerRegistry` does not require allocations to send events and can send events with references to the original packet data that is completely checked at compile time. This PR removes the events system.


TestingPlant:
This PR is no longer needed

PR #854, opened 2025-03-02
feat: add terraform setup (AI)

andrewgazelka:
I don't really know how to do this but this is what claude started with


andrewgazelka:
not important rn tbh

PR #858, opened 2025-03-05
feat: create vanilla-behaviors crate

GroobleDierne:


andrewgazelka:

this one.

PR #861, opened 2025-03-09
fix: do not multithread ChatMessage processing

TestingPlant:
Chat messages must be processed before the packet bump allocator is cleared because they store a reference to the packet data. In other words, we need a happens-before relationship where chat message processing happens before the packet bump allocator is cleared. Disabling multithreaded ChatMessage processing is one way of providing this happens-before relationship, and since processing chat messages is relatively cheap, this should not cause any noticable performance problems.


andrewgazelka:
I'm not sure I like the current way you're handling this. In an ideal world, you would have a sync point, not just making this multithreaded. I think there should be a sync point between when all of the events end and when the bomb allocator is cleared.

TestingPlant:
Replaced with #888

PR #877, opened 2025-04-12
hotfix: Bows hotifx

Tebarem:
This is just a quick fix to make bows work again without crashing and as intended.


Tebarem:
ignore this pr

PR #879, opened 2025-04-15
Fix Hyperion and bows...

Tebarem:
Some weird issues with the current Flecs_ecs so I made sure we are using a stable version and commit. Also fixed bows really quick since my old pr didnt include changes that should already have been there

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -722,20 +722,20 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.71.1"
+version = "0.70.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


TestingPlant:
I don't think we should downgrade the flecs version because we want to be able to take advantage of newer safety fixes such as https://github.com/Indra-db/Flecs-Rust/commit/ec639b3fe0d1d8ae8993e7c4178ebf66ad0d6f21 in the latest flecs main branch. The Cargo.lock file should also ensure that everyone is using the same flecs commit for compiling Hyperion so I don't believe that using the latest main branch an issue.

Tebarem:
Its a version that Indra told me to use not on hyperion but in one of my personal projects. So I decided to use it here as well and it stopped the crashes. I went through everything possible. I Rewrote blocks to use Arc<Mutex> Cause i thought the issue might be when the player spawns and couldn't get anything working it didn't make any sense. I think its just a bug with the latest commit.

PR #887, opened 2025-04-26
fix(crash): simplify attack handling logic

andrewgazelka:
- Extract reusable functions for immunity checks and combat stats calculations
- Add operator overloading via Add/Sub traits for Position and CombatStats
- Reduce nesting depth in attack handling code
- Update several dependencies including arboard and rand
- Add 'add' and 'add_assign' features to derive_more dependency

### DIFF
--- a/events/tag/src/module/attack.rs
+++ b/events/tag/src/module/attack.rs
@@ -5,7 +5,7 @@ use derive_more::with_trait::Add;
 use flecs_ecs::{
     core::{
         Builder, ComponentOrPairId, EntityView, EntityViewGet, QueryAPI, QueryBuilderImpl,
-        SystemAPI, TableIter, TermBuilderImpl, World, WorldGet, flecs,
+        SystemAPI, TableIter, TermBuilderImpl, World, WorldGet, flecs, id,
     },
     macros::{Component, system},
     prelude::Module,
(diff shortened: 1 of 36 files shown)


justfile
```diff
@@ -99,7 +99,7 @@ release-full:
         ulimit -Sn {{fds}} && cargo run --bin hyperion-proxy --profile release-full
 
 # run a given number of bots to connect to hyperion
-bots ip='127.0.0.1:25565' count='1000':
```
TestingPlant, line 102: Is this change supposed to be included in this PR?

events/tag/src/module/attack.rs
```diff
@@ -150,368 +185,128 @@ impl Module for AttackModule {
...
+
+
+                        if origin_team == target_team {
+                            let msg = "§cCannot attack teammates";
+                            let pkt_msg = play::GameMessageS2c {
+                                chat: msg.into_cow_text(),
+                                overlay: false,
+                            };
+
+                            compose.unicast(&pkt_msg, origin_connection, system).unwrap();
```
TestingPlant, line 214: Skip the attack event if it's a teammate:
```suggestion
                            compose.unicast(&pkt_msg, origin_connection, system).unwrap();
                            continue;
```

TestingPlant (changes requested):

justfile
andrewgazelka: good catch; fixed

TestingPlant (approved):

PR #888, opened 2025-04-27
feat(lifetimne): add debug and wait_no_references

TestingPlant:
wait_no_references allows the packet bump allocator to wait until there are no more references before clearing the bump allocator instead of aborting if there are any references. Debug information is also added.


andrewgazelka:
can you go a little more into this?

in which scenarios would all the references not be 0 by the end of the gametick? I could see this causing issues if people accidentally hold onto references of say one event and then many bumps (massive memory) is never cleared

TestingPlant:
> in which scenarios would all the references not be 0 by the end of the gametick? 

See #861 as an example of where this would occur. When events such as ChatMessage are processed with a multi_threaded system, the event processing threads may still be running and holding references while the clear_bump system runs and calls assert_no_references.

andrewgazelka:
> > in which scenarios would all the references not be 0 by the end of the gametick?
> 
> See #861 as an example of where this would occur. When events such as ChatMessage are processed with a multi_threaded system, the event processing threads may still be running and holding references while the clear_bump system runs and calls assert_no_references.
> 
> > I could see this causing issues if people accidentally hold onto references of say one event and then many bumps (massive memory) is never cleared
> 
> This doesn't create multiple bump allocators. It waits until all references are cleared before clearing the bump allocator. If a reference is leaked, then wait_no_references will emit warnings every 5 seconds including debug information about which reference was leaked.

ideally there should be a sync point before `assert_no_references` though where it "reads" all event related things. I think we should prio fixing sync points over this.

TestingPlant:
> ideally there should be a sync point before `assert_no_references` though where it "reads" all event related things

`wait_no_references` is a sync point; the thread is parked until all references are dropped, meaning that all events with references have also finished processing

andrewgazelka:
> > ideally there should be a sync point before `assert_no_references` though where it "reads" all event related things
> 
> `wait_no_references` is a sync point; the thread is parked until all references are dropped, meaning that all events with references have also finished processing. Once the last reference is dropped, the thread calling `wait_no_references` is unparked.

what I am saying though is why not use flecs for this natively instead of having our own custom logic? flecs has sync points natively.

TestingPlant:
This PR is no longer needed

PR #895, opened 2025-05-03
fix(ingress): don't fetch skins for players without uuids

TestingPlant:
The server can skip fetching the skin in this case because the player does not have a skin.

### DIFF
--- a/crates/hyperion/src/ingress/mod.rs
+++ b/crates/hyperion/src/ingress/mod.rs
@@ -130,21 +130,25 @@ fn process_login(
     let skins = comms.skins_tx.clone();
     let id = entity.id();
 
-    tasks.spawn(async move {
-        let skin = match PlayerSkin::from_uuid(uuid, &mojang, &skins_collection).await {
-            Ok(Some(skin)) => skin,
-            Err(e) => {
-                error!("failed to get skin {e}. Using empty skin");
-                PlayerSkin::EMPTY
-            }
-            Ok(None) => {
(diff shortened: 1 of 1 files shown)


crates/hyperion/src/ingress/mod.rs
```diff
@@ -130,21 +130,25 @@ fn process_login(
...
-                PlayerSkin::EMPTY
-            }
-            Ok(None) => {
-                error!("failed to get skin. Using empty skin");
-                PlayerSkin::EMPTY
-            }
-        };
+    if profile_id.is_none() {
+        skins.send((id, PlayerSkin::EMPTY)).unwrap();
+    } else {
```
dyc3, line 135: nit: would prefer an `if let Some(profile_id) ...` instead

dyc3 (dismissed):
Nice 👍

crates/hyperion/src/ingress/mod.rs
TestingPlant: `clippy::redundant_pattern_matching` would warn against using `if let Some(_) = profile_id` and suggest the is_some/is_none functions since we aren't using the profile id in the Some case

dyc3: Oh, interesting. I didn't see that before. Regardless, I would still prefer to have the positive condition come before the negative condition (in this case `if some else none` rather than `if none else some`), but this is a nitpick and I don't want to block you on it.

dyc3 (approved):

PR #899, opened 2025-05-17
refactor: update flecs

TestingPlant:


events/tag/src/module/vanish.rs
```diff
@@ -36,9 +36,9 @@ impl Module for VanishModule {
             &Vanished,
             &Uuid,
         )
-        .kind::<flecs::pipeline::PreStore>()
+        .kind(id::<flecs::pipeline::PreStore>())
         .each_iter(move |it, row, (compose, _connection_id, vanished, uuid)| {
-            let entity = it.entity(row);
+            let entity = it.entity(row).expect("row must be in bounds");
```
dyc3, line 41: I'm seeing a lot of changes like this. Would it make more sense to not panic?

TestingPlant: The [each_iter docs](https://github.com/Indra-db/Flecs-Rust/blob/9c7e8939a784b110c442f4f061fe46dc10fbb3cf/flecs_ecs/src/core/utility/traits/query_api.rs#L178) also uses `unwrap` to get the entity, so I don't believe there's a way to get the entity without handling the `Option`. From what I understand, `it.entity(row)` should always succeed since `row` should be a valid index. I don't believe there's a useful way to handle `it.entity(row)` returning `None` aside from panicking because it should never return `None`.

dyc3: I see, thanks for the context. So these cases would be user error and we want to surface the error as quickly as possible by panicking?

TestingPlant: `row` should always be a valid index since it's from `each_iter`. If `row` were an invalid index, that would be a bug in flecs, so it wouldn't make sense to try to gracefully handle the error within our code.

dyc3 (approved):

PR #900, opened 2025-05-18
fix: inventory slot bounds check

andrewgazelka:
## Summary
- prevent panic when mutating invalid inventory slots

## Testing
- `cargo fmt --all -- --check` *(fails: could not download toolchain)*
- `cargo test --locked --all --no-run` *(fails: could not download toolchain)*


TestingPlant:
This is a good fix. Why was it closed?

dyc3 (approved):

PR #903, opened 2025-05-20
fix: player join

TestingPlant:
This fixes the player join code to reduce the number of possible errors. The rationale of each change is explained in each commit message.


dyc3 (approved):

TestingPlant:
I found a few crashes caused by this code. I will reopen this PR once I resolve these crashes.

PR #917, opened 2025-07-08
fix: respawn

TestingPlant:
Player positions are updated immediately when TeleportEvent is sent instead of later when the player confirms the teleportation. This allows the new position to be immediately replicated to other clients, which is also useful to allow the anticheat to continue working even if the client does not acknowledge the teleportation quickly.

This also fixes a bug where a player's teleport is not synced with other players because last_tick_position is set to the current position every tick, meaning that the server will continue sending position deltas relative to the previous location instead of the full position to the teleport location.

The hyperion-respawn plugin was also removed because it is not necessary since attack.rs already handles respawn


andrewgazelka:
Closing this as unlandable, having checked each path against `main` (`68c5429`) rather than assuming.

Seven of the thirteen files this touches no longer exist, accounting for 262 of the 705 changed lines:

- `crates/hyperion-respawn/{.gitignore,Cargo.toml,README.md,src/lib.rs}` — the crate this PR deletes was already deleted, by `4608f5f` "refactor: start bedwars" (#930).
- `events/tag/{Cargo.toml,src/lib.rs,src/plugin/attack.rs}` — the `tag` event went in the same commit. `events/` now holds only `bedwars` and `smash`. `attack.rs` alone is 117 of those lines and carries most of the actual behaviour change here.

The remaining six files do still exist — `crates/hyperion/src/{egress/player_join/mod.rs,egress/sync_entity_state.rs,simulation/handlers.rs,simulation/mod.rs}`, `Cargo.toml`, `Cargo.lock` — but `simulation/mod.rs` was gutted by `96fa702` (#1016), which moved most of its contents into `flight.rs`, `lookup.rs`, `pose.rs` and `xp.rs`, so the 104 lines aimed at it are aimed at code that lives elsewhere now.

That leaves too little standing to rebase. If the respawn behaviour is still wanted, it is worth reopening as a fresh change against the current layout — the intent is not the problem, the target is. Thanks for the work, and sorry it sat long enough for the tree to move out from under it.

PR #919, opened 2025-07-16
fix(dependabot): disable on rust code

TestingPlant:
Dependabot is currently causing issues by trying to update `glam` to a new major release (0.29 -> 0.30) automatically. Dependabot is not that useful because we can always update dependencies as needed for bug fixes and new features.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,15 +3,6 @@
 
 version: 2
 updates:
-  - package-ecosystem: "cargo"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 15
-    groups:
-      all-cargo-dependencies:
-        patterns:
-          - "*"
   - package-ecosystem: "gradle"
     directory: "/extractor/"
     schedule:


dyc3:
You can tell dependabot to ignore a dependency. Why not just do that?

TestingPlant:
> You can tell dependabot to ignore a dependency. Why not just do that?

I've tried to do that here: https://github.com/hyperion-mc/hyperion/pull/911#issuecomment-3071003776

It still tried to update glam in a PR that was opened a few hours after I used the ignore command. Glam was still in the ignore conditions: https://github.com/hyperion-mc/hyperion/pull/918#issuecomment-3071582714

Should I try putting the ignore in the dependabot.yml fild? I'm not too familiar with using dependabot.

dyc3:
Weird. Must be an upstream bug.

Still, I do think that keeping deps up to date is valuable -- I've spent way too much time updating dependencies in big batches rather than gradual small updates. Over in biome we use renovate instead of dependabot, and its done pretty good. Maybe we try that instead?

TestingPlant:
I'm currently more focused on other issues, and I'm not too concerned about outdated dependencies at the moment. Feel free to help configure renovate though if you'd like. I'm not familiar with setting up either dependabot or renovate.

andrewgazelka:

this one.

PR #924, opened 2025-07-20
fix(geometry): handle when ray direction is 0

### LINKED ISSUE #909: Ray::voxel_traversal goes in the wrong direction when a component in the ray direction is exactly 0
When a component (x, y, or z) in the ray direction is `0`, `inv_direction` will contain `inf` because `1.0 / 0.0` is `inf`, and `VoxelTraversal` goes in the wrong direction. This causes the test in `hyperion/tests/entity.rs` to take a long time because, even though the arrow velocity is `(0, 1, 0)`, `Blocks::first_collision` seems to check in the z axis and only gives up once it reaches the bounds.

TestingPlant:
Fixes #909

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3044,6 +3044,7 @@ version = "0.1.0"
 dependencies = [
  "approx",
  "glam 0.29.3",
+ "itertools 0.14.0",
  "ordered-float 5.0.0",
  "rand 0.9.1",
  "serde",
--- a/crates/geometry/Cargo.toml
+++ b/crates/geometry/Cargo.toml
@@ -13,6 +13,7 @@ ordered-float = { workspace = true }
 
 [dev-dependencies]
 approx = { workspace = true }
+itertools = { workspace = true }
 rand = { workspace = true }
 tango-bench = { workspace = true }
 
--- a/crates/geometry/src/ray.rs
+++ b/crates/geometry/src/ray.rs
@@ -2,6 +2,10 @@ use std::ops::Mul;
 
 use glam::{IVec3, Vec3};
 
+const fn nan_as_inf(value: f32) -> f32 {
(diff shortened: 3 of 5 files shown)


TestingPlant:
The benchmark results are weird. I'm able to reproduce it on my machine:
```
ray_intersection/aabb_size_0.1                     [  31.2 ns ...  34.7 ns ]     +11.35%*
ray_intersection/aabb_size_1                       [  31.2 ns ...  34.8 ns ]     +11.62%*
ray_intersection/aabb_size_10                      [  31.2 ns ...  34.7 ns ]     +11.33%*
ray_intersection/ray_distance_1                    [   1.3 ns ...  13.7 ns ]    +948.09%*
ray_intersection/ray_distance_5                    [   1.3 ns ...  13.8 ns ]    +935.52%*
ray_intersection/ray_distance_20                   [   1.3 ns ...  13.8 ns ]    +935.60%*
```

Even changing `nan_as_inf` to an identity function doesn't help much:
```
ray_intersection/aabb_size_0.1                     [  31.2 ns ...  33.4 ns ]      +6.90%*
ray_intersection/aabb_size_1                       [  31.2 ns ...  33.5 ns ]      +7.42%*
ray_intersection/aabb_size_10                      [  31.2 ns ...  33.3 ns ]      +6.93%*
ray_intersection/ray_distance_1                    [   1.3 ns ...  11.7 ns ]    +773.80%*
ray_intersection/ray_distance_5                    [   1.3 ns ...  11.7 ns ]    +775.68%*
ray_intersection/ray_distance_20                   [   1.3 ns ...  11.6 ns ]    +772.15%*
```

I suspect LLVM is struggling to vectorize the code in the Vec3 impl since we're using some of the Vec3 helper methods. I'll look into it more.

PR #947, opened 2025-09-27
docs: clarify setup instructions

TestingPlant:


README.md
```diff
@@ -166,23 +166,57 @@ sequenceDiagram
...
+openssl req -nodes -newkey rsa:4096 -keyout proxy_private_key.pem -out proxy.csr -subj /
+openssl x509 -req -in proxy.csr -CA root_ca.crt -CAkey root_ca.pem -CAcreateserial -out proxy.crt -days 365 -sha256 -extfile <(printf "subjectAltName=DNS:localhost,IP:127.0.0.1")
+rm proxy.csr
+```
+
+Now run the game server. Note that the game server automatically starts the proxy.
+
+```bash
+cargo run --release --bin bedwars -- --ip 127.0.0.1 --root-ca-cert root_ca.crt --cert game.crt --private-key game_private_key.pem
+```
```
dyc3, line 202: since these are for a dev environment, maybe it would be good to include some of these long commands in the `justfile`?

dyc3 (approved):

README.md
TestingPlant: I'd like to avoid relying on external programs like `just` that people usually don't have installed so that setting up Hyperion is simpler

dyc3: I'm not saying remove these instructions necessarily, merely providing the convenience of running a simple command.

My full opinion: installing `just` is not a hurdle, its a `cargo install just` and then wait. providing an easy way to have consistent dev environments is way more valuable than the time spent debugging differences in environments, multiplied by however many people start contributing.

andrewgazelka:

this one.

PR #948, opened 2025-10-01
feat: Make server ping response data a resource

ESHARK22:
🤷‍♂️

### DIFF
--- a/crates/hyperion/src/ingress/mod.rs
+++ b/crates/hyperion/src/ingress/mod.rs
@@ -86,7 +86,7 @@ impl Default for ServerPingResponse {
 
 fn process_status_request(
     mut packets: EventReader<'_, '_, packet::status::QueryRequest>,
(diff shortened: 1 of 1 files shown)


crates/hyperion/src/ingress/mod.rs
```diff
@@ -66,8 +66,27 @@ pub fn process_handshake(
...
+                 World Record",
+            ),
+            max_players: 12_000,
+        }
+    }
+}
+
 fn process_status_request(
     mut packets: EventReader<'_, '_, packet::status::QueryRequest>,
+    ping_response_data: Res<ServerPingResponse>,
```
TestingPlant, line 89: This needs to explicitly state that there's a lifetime parameter to compile:

```suggestion
    ping_response_data: Res<'_, ServerPingResponse>,
```

TestingPlant (approved):

Issue #931, opened 2025-07-27
handle_click_slot_inner should not trust client-suggested slot changes


TestingPlant:
`handle_click_slot_inner` currently uses `packet.slot_changes` directly without verifying that they are valid. This means a malicious client can create their own items by placing them in `slot_changes` in the click slot packet.

PR #950, opened 2025-10-11
fix: Validate slot_changes in the click slot packet

Tebarem:
Fix for #931

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,3 @@ vendor/
 .trigger-debug
 .trigger
 valence/
-/certs
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ vendor/
 .trigger-debug
 .trigger
 valence/
+/certs
--- a/crates/hyperion/src/simulation/inventory.rs
+++ b/crates/hyperion/src/simulation/inventory.rs
@@ -1225,12 +1225,9 @@ fn validate_slot_changes(
             return false;
         };
 
-        // Check if slot exists in our inventory
-        if slot_idx >= inventories_mut.len() {
+        let Some(slot) = inventories_mut.get(slot_idx) else {
             return false;
-        }
-
-        let slot = &inventories_mut[slot_idx];
+        };
 
         // Skip readonly slots
         if slot.readonly {
@@ -1264,13 +1261,6 @@ fn validate_slot_changes(
                 return false;
             }
         }
-    }
-
-    // Validate that the total number of items we're trying to distribute
-    // doesn't exceed what's in the cursor
-    let total_distributed_items = i32::try_from(slot_changes.len()).unwrap();
-    if total_distributed_items > i32::from(cursor_item.count) {
-        return false;
     }
 
     true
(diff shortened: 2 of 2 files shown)


crates/hyperion/src/simulation/inventory.rs
```diff
@@ -1191,6 +1208,74 @@ fn handle_drop_key(
...
+            if slot.stack.count >= slot.stack.item.max_stack() {
+                return false;
+            }
+        }
+    }
+
+    // Validate that the total number of items we're trying to distribute
+    // doesn't exceed what's in the cursor
+    let total_distributed_items = i32::try_from(slot_changes.len()).unwrap();
+    if total_distributed_items > i32::from(cursor_item.count) {
```
TestingPlant: I'd recommend changing this to `==` so players can't destroy their own items
```suggestion
    if total_distributed_items == i32::from(cursor_item.count) {
```

.gitignore
```diff
@@ -34,3 +34,4 @@ vendor/
 .trigger-debug
 .trigger
 valence/
+/certs
```
TestingPlant: This should be in a separate PR. I'm thinking of adding it to #947 and ignoring all `.crt`, `.pem`, and `.csr` files

crates/hyperion/src/simulation/inventory.rs
```diff
@@ -1191,6 +1208,74 @@ fn handle_drop_key(
...
+            // For non-empty slots, they should have space available
+            if slot.stack.count >= slot.stack.item.max_stack() {
+                return false;
+            }
+        }
+    }
+
+    // Validate that the total number of items we're trying to distribute
+    // doesn't exceed what's in the cursor
+    let total_distributed_items = i32::try_from(slot_changes.len()).unwrap();
```
TestingPlant: Wouldn't this only track the number of slots changed, not the total number of items?

crates/hyperion/src/simulation/inventory.rs
```diff
@@ -1191,6 +1208,74 @@ fn handle_drop_key(
...
+        let Ok(slot_idx) = usize::try_from(slot_change.idx) else {
+            return false;
+        };
+
+        // Check if slot exists in our inventory
+        if slot_idx >= inventories_mut.len() {
+            return false;
+        }
+
+        let slot = &inventories_mut[slot_idx];
```
TestingPlant: The manual bounds check can be removed here:
```suggestion
        let Some(slot) = inventories_mut.get(slot_idx) else {
            return false;
        }
```

TestingPlant (changes requested):

crates/hyperion/src/simulation/inventory.rs
Tebarem: If you have 32 diamonds in cursor and drag over 6 slots, you'd distribute ~5 per slot with 2 remaining. With ==, this would be rejected because 6 != 32

.gitignore
Tebarem: Alright. Just had it quickly setup so I don't accidentally commit certs

crates/hyperion/src/simulation/inventory.rs
Tebarem: You're right. I think we already validate everything correctly so I'm just going to remove it.

TestingPlant (approved):

andrewgazelka:

this one.

PR #956, opened 2026-07-27
refactor: revert the ECS to flecs, modernised to flecs_ecs 0.2.2

andrewgazelka:
Reverts the ECS from `bevy_ecs` back to flecs, modernised to `flecs_ecs 0.2.2`, forward-porting the 16 commits that landed on top of the original migration.

Spine is `git revert 4f2bafe` — the single commit that moved flecs → bevy in July 2025 (149 files, +9598/−9026) — then reconciling everything since.

## Why the API work was more than a revert

The old code pinned `flecs_ecs 0.1.3`. **Every 0.1.x is now yanked**; 0.2.2 is the only installable release. The delta that mattered:

- The `($)` singleton DSL and term-level `.singleton()` were removed. The replacement is the flecs 4.1.1 component trait — `world.component::<T>().add_trait::<flecs::Singleton>()` once, and queries match automatically. That was ~55 call sites across 20 files, mostly deleted rather than rewritten.
- `id::<T>()` → `T::id()`, and bare types inside relationship pairs.
- `TableIter::entity(i)` no longer returns `Option`.
- `.multi_threaded()` on the system builder is gone, replaced by the `ParSystemAPI` callbacks (`par_each`, `par_each_entity`, …) which set the flag themselves.

## Two things reviewers should look at

**Play packets no longer decode in parallel.** flecs 4.1.0 forbids creating entities inside a multithreaded system, and play handlers spawn entities, so play decode/dispatch is a single-threaded `each_iter`. Handshake, status and login keep their parallel decode. This is pre-bevy flecs behaviour but a throughput regression against bevy HEAD on the hottest path, and worth a benchmark before the next event.

**Packet ordering stayed deleted.** The revert wanted to restore `crates/system-order` and its ~40 call sites — the machinery that queried the `DependsOn` graph for each system's topological depth and packed it into a 16-bit ordering key. But ordering was deliberately removed in `b5cc17c` ("added extra complexity which is no longer needed"), so the channel-based approach from that commit is kept and every restoration point was suppressed.

## Toolchain

`nightly-2025-02-22` → `nightly-2025-05-05`. `flecs_ecs 0.2.2` declares `rust-version = 1.88` and `flecs_ecs_derive` uses unguarded let-chains, so the old pin cannot build it. 2025-05-05 is the narrowest bump that works — 2025-06-26 was tried first and rejected, since it breaks the `bvh` git dependency with an E0391 cycle in code this change never touches.

## Deliberate deviations from a straight revert

- **The networking transport stays at HEAD.** `packet-channel`, the packet-state decode pipeline and `hyperion-packet-macros` are transport, not ECS, and every post-bevy proxy commit builds on them.
- **`RuntimeLifetime`/`LifetimeHandle` deleted.** valence's `feat-bytes` removed the lifetime from `Decode`, so decoded packets own their payloads. Keeping the borrow-extension machinery would have been unsound, not merely redundant — `ChatMessage.msg` outlives its handler. `packet_switch` is no longer `unsafe` as a result.
- **`CommandChannel` reimplemented** as a queue of `FnOnce(&World)` drained in `OnLoad`. bevy's was a type-erased byte buffer; async proxy tasks need the same deferral either way.
- **`Compose` no longer takes `&World`** — it uses the `thread_local` crate rather than the stage-indexed thread-local.
- `simulation/packet_state.rs` is gone; flecs models protocol state as an exclusive enum relationship. `crates/hyperion/tests/packet_state.rs` pins that, and the guard was watched failing before being trusted.

## Verification

```
cargo clippy --all-targets --all-features -- -D warnings    rc=0
nix run .#test                    92 tests run, 92 passed
nix run .#fmt -- --check          rc=0
nix run .#doc                     rc=0
rg -n bevy --glob '**/Cargo.toml'  no matches
```

Merged with main after #955; the only conflict-adjacent fix was a stale `#[expect(clippy::missing_const_for_fn)]` that the newer nightly reports as unfulfilled.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andrewgazelka:
Note on `compare (general)`: it fails with SIGILL, and the cause is pre-existing rather than anything in this branch. `bench.yml` built with `-Ctarget-cpu=native`, and the compare job downloads a baseline binary the export job built on a *different* runner and executes it. Hosted runners are not a uniform CPU pool, so the baseline can hit instructions the compare host lacks. It has always been a coin flip; this branch lost it.

Fixed here by building both sides for `x86-64-v3`. The job will keep failing on this PR because the baseline it downloads was built before the fix — it clears once main exports one built the new way.

Also fixed in passing: `bench.yml` hardcoded `nightly-2025-02-22` rather than reading `rust-toolchain.toml`, so it could not compile `flecs_ecs 0.2.2` at all; and the compare step redirected stdout to a file, so a failure showed up as a bare exit code with nothing to diagnose it.

PR #958, opened 2026-07-27
feat: hot reload flecs modules, refusing layout changes without a migration

andrewgazelka:
Hot reloading for flecs game modules, with **schema migrations** as the mechanism: edit Rust, see it live without restarting; and when a component's layout changes incompatibly, the reload is **refused** unless a migration exists.

## Refusing is the point

Nearly all prior art either forbids layout changes or corrupts state silently on one:

| | behaviour on a layout change |
| --- | --- |
| `hot-lib-reloader` | README: UB. Names our exact case — ECS `TypeId` mappings break |
| `dexterous_developer` (archived 2025-05) | `rmp_serde` round-trip; on failure the entity returns **without the component**, a resource falls back to `Default`. Silent partial data loss |
| Dioxus `subsecond` | Declines outright: "does not support hot-reloading of structs… the program will crash" |
| Unreal Live Coding | Reinstancing "often silently leave[s] affected blueprint assets in a corrupted state with zero warnings" |
| MSVC Edit-and-Continue | Refuses, with a diagnostic — the model followed here |

flecs itself has no unload counterpart to `ecs_import`, but `ecs_abort(ECS_INVALID_COMPONENT_SIZE)` on re-registering at a different size is always on, even in release. That dictated the design.

## How it works

The manifest is read from a **live world** via flecs meta reflection rather than generated at build time. Component identity across a reload is the flecs **symbol** — ids churn every reload and `TypeId` has no cross-artifact guarantee. A new build registers into a throwaway **probe world** and is diffed there; the live world is only touched once the diff is accepted.

A migration is four lines, and the declared old layout is checked against the world before a byte is read:

```rust
hyperion_hot_reload::migration! {
    component: Health,
    from: { hp: u32 },
    with: |old| Health { hp: old.hp as f32 },
}
```

## What the detector catches and misses

**Catches:** primitive type change at equal size (`u32`→`f32`), any size or alignment change, fields added, removed, renamed or reordered (real `offset_of!` offsets, so compiler reordering too), nested interior changes to depth 16, enum variant add/remove/rename/reorder, unreflected components holding data, migrations that misdescribe the old layout, wrong-rustc and statically-linked modules.

**Misses, stated plainly:** semantic change at identical layout, explicit discriminant values, pointers and handles that reflect as clean `uptr` leaves, module-local statics. Data-carrying enums are **untested** and said to be untested rather than guessed at.

Anything not reducible to primitive leaves makes the whole component `Unknown` and is refused. Partial checking is the dangerous case.

## Three defects found in the first attempt

1. **Identity was the entity path, not the symbol.** `world.import::<M>()` scopes components under the module entity, so `Health` registers at `::demo::ArenaModule::Health` while its symbol stays `demo::Health`. Lookups never matched and the component tree came back empty on every load.
2. **Nested structs were recorded by type name.** Change `Inner{a:u16,b:u16}` → `{a:u8,b:u8,c:u16}` and size, alignment and the enclosing member list are all identical — accepted silently, wrong at every byte past the first field. Now flattened to primitive leaves at absolute offsets.
3. **Migration field types used `flecs::meta::u32`**; flecs' real symbol is bare `u32`, so every migration would have been rejected as mismatched.

Plus two flecs behaviours that would have broken it quietly: `World::import` early-returns if the module entity carries `flecs::Module`, so a reload would delete the old systems and never create new ones — but *deleting* the module entity cascades through `(ChildOf, OnDeleteTarget, Delete)` and takes its components' data with it. The tag is cleared and the entity kept.

## The one nearly shipped

The demo printed four correct transcripts and exited **139**. Invisible because the output was piped to `tail`. `libloading::Library::drop` calls `dlclose`, and `HotReloader` dropped before `World`, unmapping images that teardown then called into. The code carried a comment claiming the libraries were "deliberately never closed" while doing the opposite.

## All four guards watched failing

| Guard | Broken by | Observed |
| --- | --- | --- |
| Gate refusal | comparing leaves by offset/size only | `kept: ["demo::Health"]` — the exact silent reinterpretation |
| Flattening | recording nested members by name | both structs → `[(0,4),(4,4)]`, identical |
| Registry guard | a direct default-featured `flecs_ecs_sys` dep | exit 1, naming `regenerate_binding` |
| ABI anchor | module built without `-C prefer-dynamic` | refused, naming both anchor addresses |

## Demonstration

`demo.sh`, rc=0, one world with three entities created once and never rebuilt:

- **code-only change** → ACCEPTED, `instances rewritten: 0`, hp 11/21/31 → 16/26/36 (regen +1 became +5, so new code ran), `points` untouched
- **`u32`→`f32`, no migration** → REFUSED, naming the component, both schemas and a paste-ready stub; the world kept ticking on the old build
- **same change with the migration** → ACCEPTED, 3 instances rewritten, 21 → **22.0**. A raw reinterpretation of `21u32` would read `2.9e-44`

## Caveats

No file watcher, no rollback past teardown, dylibs leak unboundedly, **verified on aarch64-darwin only** and the ELF branch of `build.rs` is written but unexercised. Not wired into hyperion's own modules — no existing crate is touched, and `docs/hot-reload.md` carries the exact diff they would need instead.

**18 of 111 `derive(Component)` structs currently carry `#[flecs(meta)]`, so 84% of real components would be refused today** as unprovable. That is the honest gap between this and production use.

## Verification

```
nix run .#test    107 tests run, 107 passed
nix run .#lint    rc=0
nix run .#fmt -- --check   rc=0
nix run .#doc     rc=0
./crates/hyperion-hot-reload/demo.sh   rc=0
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andrewgazelka:
Force-pushed: the branch was carrying an earlier, defective attempt at this rather than the corrected one.

A first pass at this design died partway through on an unrelated API error. A second pass salvaged it and found three defects that made the original non-functional or unsafe:

1. **Identity was the entity path, not the flecs symbol.** `world.import::<M>()` scopes components under the module entity, so `Health` registers at `::demo::ArenaModule::Health` while its symbol stays `demo::Health`. Lookups never matched, so the component tree came back empty on every load — which is what killed the first run.
2. **Nested structs were recorded by type name.** Change `Inner{a:u16,b:u16}` to `{a:u8,b:u8,c:u16}` and the size, alignment and enclosing member list are all identical, so it would be accepted silently and wrong at every byte past the first field. Now flattened to primitive leaves at absolute offsets.
3. **Migration field types used `flecs::meta::u32`**; flecs spells it bare `u32`, so every migration would have been rejected as mismatched.

The branch now holds the corrected version. Verified on it: `nix run .#test` 107 passed, `.#lint`, `.#fmt -- --check` and `.#doc` all rc=0, and `demo.sh` rc=0 with all three cases — code-only swap keeping state, `u32`→`f32` refused, then accepted with a migration (21 → 22.0, where a raw reinterpretation would read 2.9e-44).

PR #977, opened 2026-07-27
fix: remove the packet inspector, the last pin on a second valence

andrewgazelka:
`nix flake check` fails on main before it builds anything:

```
error: Cargo.lock contains multiple git dependencies with the same name-version:
valence_protocol-0.2.0-alpha.1+mc.1.20.1, valence_nbt-0.8.0, ...
cargo-unit cannot generate an aggregate vendor dir for this lock without losing source identity.
```

The whole workspace pins valence on `feat-bytes`. `tools/packet-inspector` alone pinned `feat-open`, and that is the entire duplicate.

## Why not just repin it

Tried first. It does not build against `feat-bytes`:

```
error[E0432]: unresolved import `valence_protocol::decode`
error[E0107]: trait takes 0 lifetime arguments but 1 lifetime argument was supplied
error[E0277]: the trait bound `HandshakeC2s<'_>: Decode` is not satisfied
```

So it has been compiling only because it kept its own copy of the library.

## Why remove rather than port

It is a GUI proxy that decodes 1.20.1 packets, which is the protocol this repo is leaving, and `tools/client-26.2.py` already reads the wire for 776. A 776 inspector would be a rewrite against `hyperion-minecraft-proto`, not a repin, and it would not need egui, eframe and wgpu. One `git revert` brings this back if that is wanted.

`libvoidstar.so` goes too: 363KB of Antithesis instrumentation with no reference left after #976 removed the bot that loaded it.

Also carries the `mainProgram` fix for the vanilla encoder. Without it `lib.getExe` guesses `bin/minecraft-vanilla-encoder`, and with `abort-on-warn` set that warning is itself a flake check failure.

## What this does not fix

`nix flake check` now gets past evaluation and reaches the unit graph, where it hits a different fault:

```
local unit bedwars 0.1.0 source root /nix/store/5av5x4gg...-s5hs2qnl...-source/events/bedwars
is outside workspace root /nix/store/s5hs2qnl...-source
```

Note the doubled store hash in the member path. That is in cargo-unit, not here. Observed on aarch64-darwin; this PR's CI will say whether linux hits it too.


andrewgazelka:
The `Flake` red left after this PR is now filed upstream as [indexable-inc/index#4239](https://github.com/indexable-inc/index/issues/4239), with a fix in progress.

Diagnosis: `cargoUnit.buildWorkspace` ties each local unit's package root back to `workspaceRoot` only, but the unit graph is planned in a content-stubbed tree derived from `src`. When a caller passes `src = ./.` and `workspaceRoot = ./.`, which is the shape cargo-unit's own docstring recommends, those are two different store paths and every local member fails the tie-back. Reproduces on both aarch64-darwin and x86_64-linux, and on index `main` as well as the pinned rev, so it is not a stale pin and it is not this PR.

Merging on the two reds this PR does clear: `unused-deps` and the duplicate valence source.

PR #978, opened 2026-07-27
feat(net): serve the play state at protocol 776

andrewgazelka:
A 26.2 client reached play state on `main` and then received 1.20.1 packet numbers. Measured with `tools/client-26.2.py`:

```
0x1E DebugSample                  x4096
0x24 Explode                      x3682
RESULT: 0 LevelChunkWithLight packet(s)
```

0x24 was chunk data in 1.20.1 and means Explode in 776; 0x1E was chunk unload and means DebugSample. The client saw no terrain.

After this branch, same command:

```
0x0B ChunkBatchFinished           x257
0x0C ChunkBatchStart              x257
0x25 ForgetLevelChunk             x4096
0x2D LevelChunkWithLight          x4096
0x65 SetEntityMotion              x269
RESULT: 4096 LevelChunkWithLight packet(s)
RESULT: GameEvent(LEVEL_CHUNKS_LOAD_START) sent
```

## The gate

`nix run .#e2e` is new. It boots the stack on its own ports, drives that client through a real join, and exits with its verdict. It is the only check that reads the wire; every other gate reads the source, and a test proving a packet encodes to the right bytes stays green while the server sends it under another version's number.

The client's own check changed with it. Testing that every id decodes as a valid 776 id passes a 1.20.1 packet by accident. It now tests meaning: a server that has just accepted a login has no reason to detonate anything, so an Explode is proof the numbering is wrong. Watched it fail with that message before trusting it.

## Five ports

**Terrain.** Chunk data, unload, block updates, batch bracketing. Block state ids are translated 1.20.1 to 26.2 through names at startup, from a table generated out of the jar: 1,196 rows describing all 32,366 states, because ids are contiguous per block and the states inside a run are a mixed-radix product of the properties. 12 blocks enumerate their properties in a different order than they declare them. Two Mojang renames (`grass` to `short_grass`, `chain` to `iron_chain`) and a test over all 24,135 states so a version bump fails in CI rather than turning the world to air. An unmapped state panics naming the block.

**Entities.** Two corrections from the decompiled source rather than the wiki: `AddEntity` carries velocity through `Vec3.LP_STREAM_CODEC`, not three shorts, and the field order moved; and `EntityPositionSync` rather than `TeleportEntity` is the observer-facing teleport. Tracked-data indices were recovered by reflection over the pinned jar, which found that `Avatar` now sits between `LivingEntity` and `Player` and shifts absorption and score, and that the old table was already wrong for 1.20.1 on displays.

**Player-facing.** Tab list, chat, sound, health, abilities, command tree, titles, scoreboards. Reference bytes captured from Mojang's own `StreamCodec`s by compiling a probe against the server jar.

**Serverbound.** 20 handled, 14 ignored with the reason on each arm, 35 logged loudly. Three behaviour changes: attack split from interact; sneak moved to `PlayerInput`, which the 763 path never read, so a crouching player kept a standing hitbox; movement ends in a flags byte, and reading it as a bool calls an airborne player scraping a wall grounded.

**Inventory.** Item ids resolve by name, never by number: a diamond sword is 797 in 1.20.1 and 964 in 26.2, and the menu registry shifted because 26.2 inserted `crafter_3x3`. The cursor moved to `SetCursorItem`; the 1.20.1 spelling still parses and is now dropped.

## Also here

- `Writer::from_vec`/`clear` and one encode buffer per thread instead of one per packet. The play state sends a packet per entity per tick per viewer; the comment on that impl already said a play port would want this.
- Tick time, warm, one player streaming chunks: 1.04 ms on the 763 path, 1.01 ms on this one. An earlier version measured 1.56 ms because `PalettedContainer::from_values` seeds palette id 0 with its `default` whether or not it appears, and seeding air pushed a 16-block section to 17 entries and across the four-to-five-bit boundary.

## Verified

```
cargo check --workspace --all-targets   rc=0
nix run .#lint                          rc=0
nix run .#test                          rc=0   435 passed, 1 skipped
nix run .#e2e                           rc=0
```

## Not done, and it is a real list

25 valence play packets remain, none on the join path. Another player appearing (PlayerSpawn, EntitySpawn, EntitiesDestroy), combat (EntityDamage, DamageTilt, DeathMessage, EntityStatus, EntityAttributes, EntityVelocityUpdate), scoreboards and the tab list header, chat and titles from game code, and eight others. 89 references to valence remain in the server crate.

Spawning specifically is blocked on `EntityKind`, whose discriminants are 1.20.1 registry ids with no names attached. A generated table for that is in flight.

Block entities go out as an empty list: their type ids are a different registry in 776 and their tags predate data components, so a chest keeps its shape and holds nothing. Sending them untranslated would disconnect the client. Heightmaps are still the placeholder the 763 path sent.

No vanilla client has connected yet. Everything above is a scripted client, which proves the numbering and the counts and nothing about how it looks.


andrewgazelka:
`Flake` is the one red. It is pre-existing on main and now diagnosed and fixed upstream: [indexable-inc/index#4242](https://github.com/indexable-inc/index/pull/4242) for [#4239](https://github.com/indexable-inc/index/issues/4239).

Cause turned out not to be the tie-back rule at all. cargo-unit's render derivation spelled `args.src` two different ways in one builder script, interpolated in a `sed`, stringified in the `--workspace-root` argument. Those agree only when `src` is a derivation, which is what every existing caller and every existing test passed. For a path value the interpolated form gets a second store hash, which is the doubled path in the error. With that branch overridden in, hyperion's `nix flake check` gets past the unit graph with zero occurrences of the error.

Merging on everything else green: Clippy, Documentation, Tests, deny, unused-deps, Code Coverage, Formatting, plus locally `nix run .#e2e` rc=0.

PR #985, opened 2026-07-27
fix(nix): make the vanilla encoder run against 26.2, and its output reproducible

andrewgazelka:
`nix flake check` was down to one failing check. This makes it green.

## What was wrong

The harness that prints reference bytes from the game's own encoders threw before writing anything:

```
java.lang.NullPointerException: Components not bound yet
  at net.minecraft.core.Holder$Reference.components(Holder.java:281)
  at net.minecraft.world.item.ItemStack.<init>(ItemStack.java:265)
  at VanillaEncoder.setEntityData(VanillaEncoder.java:408)
```

`Bootstrap.bootStrap()` used to be enough. In 26.2 it is not.

## What binds them, and how that was found

Grepping 7,445 extracted classes for `bindComponents` named exactly two: `Holder$Reference`, which defines it, and `DataComponentInitializers$BakedEntry`, the only caller. Walking back with `javap -c -p`: `BakedEntry.apply` is called from `PendingComponents.apply`, from `ReloadableServerResources.updateComponentsAndStaticRegistryTags`. `Bootstrap.bootStrap`'s bytecode calls `BuiltInRegistries.bootStrap` and `CreativeModeTabs.validate` and stops. It never touches any of it.

`javap -c WorldLoader` gave the sequence a real server runs, which the harness now mirrors: open the vanilla pack, load tags for existing registries, build updated lookups, `RegistryDataLoader.load`, then apply pending tags and pending components.

That also cleared the two failures queued behind it, both of which needed datapack-loaded registries.

## Fixture values that changed

Four, plus 82 additions, which are the play packets the committed file never had because the harness died before writing them.

`biome_id.plains` 29 to 40, `biome_id.desert` 63 to 14, and the two palette fixtures that follow them. The old ids came from `listElements` on a `createLookup` provider. 40 and 14 are the indices of `plains.json` and `desert.json` in the pack's sorted listing, which is the order `RegistryDataLoader` assigns, and they now agree with the generated biome table, which they did not before.

## A reproducibility bug found on the way

The registry dump was not reproducible, in two independent ways, and the check only passed because nix had the output cached. Any edit to the harness reveals it.

Element order came from that same provider, which is a `HashMap` over freshly allocated holders. The payload bytes of two registry blobs came from `DimensionType.attributes` being keyed on identity hash codes, which leaks through `CompoundTag`'s `HashMap` because it keeps insertion order within a bucket. Fixed by reading a real registry and sorting compound keys. Two runs are now byte-identical, and all 24 blobs are regenerated.

Also removed: a `catch` that turned an encode failure into a `skipped.json` entry with a zero exit. It was watched swallowing a real failure and producing a 14-of-29 dump that looked fine.

## Test vectors moved onto the harness

Two files' worth of hand-derived bytes are now Mojang's own.

- `tests/play_entity.rs`, fully. Two literals remain and both are inputs rather than expectations.
- `tests/play_combat.rs`. Every vector matched byte for byte except one: `AttributeModifier.id` is an identifier, so a server writes `minecraft:hi` where the hand-derived vector had a bare `hi`. The Rust side was wrong and is fixed.

One left behind, documented in place: `ClientboundEntityEventPacket`'s only public constructor takes a live `Entity`, which needs a `Level`, which needs a server.

## Verified

```
nix build .#minecraft-encoder-fixtures   rc=0
nix flake check                          rc=0   all checks passed
nix run .#lint                           rc=0
nix run .#test                           rc=0   447 passed, 1 skipped
```


andrewgazelka:
The remaining `Flake` red is the CI runner, not this branch. Locally on this exact commit:

```
$ nix flake check --accept-flake-config
all checks passed!
rc=0
```

CI fails instead with a store path that does not exist, preceded by the daemon refusing `ca-derivations`, and two runs of the same commit named two unrelated paths. hyperion is built content-addressed throughout, so a daemon without that feature cannot realise most of the graph and the symptom surfaces one derivation at a time. Filed as ENG-10485.

PR #1005, opened 2026-07-28
feat: one identity per connection, adventure mode, and a kit mob on your back

andrewgazelka:
Three things a player notices, and one protocol finding that shapes all of them.

## Two people with the same IGN can both join

The profile id was a hash of the username, so two connections calling themselves
Steve were one player to every client: `playerInfoMap` is keyed on profile id and
filled with `putIfAbsent`, so the second arrival was dropped from the tab list,
rendered with the first one's profile, and took the first one's entry with them
when they left. The id is now random per connection.

**The name is left exactly as typed, and that is a decision.** Nothing on the
wire is keyed on a name: the tab list, `AddEntity` and `PlayerInfoRemove` all
carry the profile id, the sidebar and the death messages read the flecs entity
name, and kits are looked up by component rather than by path. A suffix would put
a string on four screens that the player did not type, to solve a problem none of
those four surfaces has. The cost is that an offline id no longer survives a
reconnect, and nothing wanted that: the only thing keyed on a profile id is the
Mojang skin cache, consulted only for real online-mode profiles.

Writing the gate turned up an older half of the same bug. Every player was told
one profile id in `LoginFinished` and every other client a different one, because
an observer in `SimModule` hands any entity gaining an `EntityKind` without a
`Uuid` a fresh random one. The login handler builds the player from inside a
system, so its `set` is queued; the observer fires while an earlier queued command
merges, sees no `Uuid` yet, and its own write is appended last and wins. A player
was therefore never `isLocalPlayer` to their own client. The observer now skips
anything holding a `ConnectionId`. ENG-10813.

## Players are in adventure mode, and cannot break the arena

Gamemode is a server capability rather than something each event improvises: a
flecs enum on the player, exclusive so a new mode drops the old one, a
`DefaultGamemode` singleton every joining player is put into, and one observer
that keeps the client's abilities and the tab list icon in step. The block
handlers ask `may_build()` before turning a packet into an event, so a client that
is not vanilla gets the same answer as one that is. smash sets the default to
adventure and its spectator path now sets the component instead of writing a
`GameEvent` by hand.

## A player wears their kit's mob

`KitSkin` lives on the kit prefab and is reached through the `(Playing, kit)`
relation that already says which mob you are, so there is one copy of the answer
and it is on the kit. The adapter is the only thing that copies it across, because
hyperion speaks profiles and the game half does not know hyperion exists.

### What the client enforces at protocol 776, and where that was checked

Read out of the 26.2 client jar, sha1 `2dc72797acbc1b63fc16a11c4ac393605f453754`,
which is `downloads.client` in the 26.2 version manifest. **26.2 ships
unobfuscated**, so these are Mojang's own names rather than a mapping's guess.

- `PlayerInfo.createSkinLookup` builds the lookup with `requireSecure =
  !minecraft.isLocalPlayer(profile.id())`.
- `SkinManager.createLookup` keeps a skin only when `!requireSecure ||
  skin.secure()`, otherwise falling back to `DefaultPlayerSkin`.
- `skin.secure()` is `textures.signatureState() == SignatureState.SIGNED`.
- authlib 9.0.75 `getPropertySignatureState` returns `UNSIGNED` with no signature
  and `INVALID` when no Mojang profile property key validates it.
- `unpackTextures` discards the whole payload unless every texture URL passes
  `TextureUrlChecker`, whose allow-list is exactly `{"textures.minecraft.net"}`.

So: an unsigned `textures` property dresses a player up **for themselves and for
nobody else**; none of it is conditional on the server's online-mode flag; and a
skin cannot be self-hosted. The player-info path does carry a skin after join, and
that is what makes a kit change visible, because the client caches a `PlayerInfo`
per profile and adds with `putIfAbsent`, so the profile has to be removed and
re-sent.

That failure has no server-side symptom, so it needs a gate rather than care.
Every kit's skin is a real Mojang-signed payload and `nix run .#check-kit-skins`
verifies each signature offline against Mojang's committed public keys, the way
`YggdrasilServicesKeyInfo.validateProperty` does.

## What is not checked

**No human has looked at the fifteen skin images.** Selection was mechanical:
MineSkin's search already filters by mob name and tag, and within that pool the
winner is the skin whose colour histogram is closest to the mob's actual vanilla
texture out of the client jar. Distances and provenance are tabulated in
`events/smash/skins/README.md`. Palette distance is weakest where a mob is near
monochrome, and the four worst matches are guardian, spider, sky_squid and wolf.
Swapping one is a two-file change with no code in it.

## Gates

`nix run .#smash-identity-e2e` drives two real connections through all of it and
is a `nix flake check` derivation. Nine assertions, all green:

```
PASS both connections announcing the IGN 'Twin' finished login
PASS the two share an IGN and have different profile ids (b4c7e8bf..., e439ac17...)
PASS each client's tab list holds both players as separate entries
PASS first sees the name exactly as typed, unsuffixed: ['Twin']
PASS second sees the name exactly as typed, unsuffixed: ['Twin']
PASS the server publishes adventure mode for both players: {'first': 2, 'second': 2}
PASS the server acked the dig sequence, so the client rolls its prediction back: [1, 1]
PASS digging (0, 64, 6) did not change the block: []
PASS the profile the other player sees carries a textures property
PASS that texture is the one committed for Zombie
PASS and it carries its Mojang signature, without which only the wearer would see it
RESULT: ok (0 checks failed)
```

Both guards were watched failing before being trusted: a tampered signature, an
empty signature and a kit declaring no skin each fail `check-kit-skins` with the
message intended, and the identity gate failed on all four of the real bugs above
before they were fixed.

`nix run .#test` is 585 passed, 1 skipped.

Writing the gate also caught a claim the code made and did not honour.
`ConfirmBlockSequences` is pushed to by two handlers and drained by nothing, so a
refusal that "confirmed the sequence" acknowledged nothing and left the client
showing a hole in the arena. Refusals now send the `BlockChangedAck` directly, and
the inert queue is labelled. ENG-10806.

The bedwars skin module was a private copy of the profile refresh that sent the
literal username "Player" and no gamemode, which is how the two drifted. Deleted
in favour of the shared path.

---

AI-authored, by Claude. No human has reviewed this yet.


andrewgazelka:
## Update after rebase onto 8c04c38

Rebased onto current main (#1002 tab completion, #1004 text API, #1003 sounds).
Conflicts were additive and resolved keeping both sides: the kit builder now has
both `.sounds(..)` and `.skin(..)`, and the adapter imports both seams.

Local gates re-run on that base, exit codes read from the logs rather than from a
notification:

- `nix run .#smash-identity-e2e` rc=0, eleven assertions, `RESULT: ok (0 checks failed)`
- `nix run .#test` rc=0, 608 passed 1 skipped
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` rc=0
- `nix run .#check-kit-skins` rc=0, 15 kits

## One red that is not this branch

`nix run .#smash-e2e` fails on main with `NOT PROVED cooldowns refused a second
use`, one COOLDOWN FAILED line per ability. Filed as ENG-10833.

It is #1004's, not this branch's. #1004 turned the refusal into a coloured
component, and `tools/smash-match.py`'s `take_nbt_string` reads only a bare
`Tag::String` and says in its own docstring that a bare string is "every message
this server sends" -- which stopped being true in that commit. Confirmed by
running the gate in a scratch worktree at plain `origin/main` with none of this
branch's code, where it fails identically. The same branch rebased onto 162dc01,
before #1004, passed all ten steps.

Not fixed here: it belongs to the text seam rather than to identity.

PR #1008, opened 2026-07-28
test: fire drill for the ENG-10817 flake gate, not for merge

andrewgazelka:
Not for merge. ENG-10817 fire drill: watching the newly enforced gate fail.

This PR breaks the game server module argv (swaps --port ahead of --ip) so that
checks.nixos-modules must fail. Every other CI job reads rust and cannot see
it. The point is to confirm the Flake job goes red with the message intended,
rather than trusting a guard nobody has watched fire.

Closed and the branch reverted as soon as CI reports.

Authored by an AI agent (Claude Opus 4.5 via Claude Code).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,13 @@ on:
 jobs:
   # Each check is the same flake app a contributor runs locally, so CI cannot
   # drift from `nix run .#ci`.
-  # Job names are fixed by the merge queue ruleset on main, which requires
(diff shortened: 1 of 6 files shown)


andrewgazelka:
Fire drill done. It went red with the message intended (job 90222622214): FAILED nixos-modules, hyperion NixOS module argv lost: /bin/bedwars --ip :: --port 35565. Closing unmerged; the break is reverted on main by never landing.

PR #1055, opened 2026-07-28
test(smash): reusable packet monitor + skin/hat regression gate

andrewgazelka:
## What

A controllable headless client is already in the repo (the scripted protocol-776 Python clients driven by `nix/e2e.nix`); azalea cannot help here (its latest is mc26.1, this server is 26.2/776) and `rust-mc-bot` is stuck at 763. So this extends that harness rather than adding a new client:

- **`tools/packet_monitor.py`** - one reusable decoder + accumulator for the two clientbound packets a skin or visual check reads:
  - `PlayerInfoUpdate` (0x46): the tab-list profile, including the `textures` property and its Mojang signature.
  - `SetEntityData` (0x63): entity metadata, including the byte at **index 16** (`Avatar.DATA_PLAYER_MODE_CUSTOMISATION` on 26.2) whose bits are the second skin layer - cape, jacket, sleeves, trouser legs and the **hat**.
  - `AddEntity` (0x01): the uuid->entity-id map that ties a profile to the body whose metadata carries its overlay mask.
  - It replaces two drifting copies of `parse_player_info_update` that lived in `identity-check.py` and `smash-selector.py`; both now import it.
- **`tools/skin-check.py` + the `smash-skin-e2e` gate** - drives two real clients: one puts on a kit, and the other's view is asserted to carry the committed **signed** skin AND the overlay mask with the hat bit set, and that putting on a kit did not Respawn the wearer.

The harness is store-cached and deterministic: the client files ride the existing `lib.fileset.toSource` + `mkCheck` plumbing, keyed on content.

## Reproduce -> fix -> prove (the skin and hat bugs)

Captured against a running smash server. On 26.2 **the wire is already correct for every other player**: another player sees the committed signed textures and a `0x7F` overlay mask (hat included), with no Respawn. The historical "no hats" bug (`pre_play.rs:683`: unhandled `ClientInformation` left the mask at zero) is fixed but had no regression test - this gate is that test.

Guard proven fail-then-pass against the live server:
- committed source: all 8 checks green;
- break `roster::entry_of` (drop the property) and `metadata::show_all` (zero the mask): 5 checks red, gate exits non-zero;
- restore: green again.

## Not in this PR

The one symptom that is NOT explained by the wire is "a player sees their own skin, not the kit's". That is the **local player's own avatar** - a 26.2 client builds `LocalPlayer` at login from its own account skin and does not rebuild it from a tab-list update, so the wearer keeps their login skin while everyone else sees the kit. `crates/hyperion/src/egress/player_join/roster.rs` documents this tradeoff and deliberately leaves it (the only rebuild path found is a Respawn, which throws the wearer's world away). It needs operator confirmation of the exact case and the 26.2 client decompile to check for a lighter refresh path; tracked separately.

## CI

hyperion CI is known-broken; validated locally against a running server (`nix run .#smash-skin-e2e`, `smash-identity-e2e`, `smash-selector-e2e` all green after the dedup).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -381,6 +381,7 @@
             completions-e2e = 6000;
             smash-hotbar-e2e = 7000;
             smash-hud-e2e = 8000;
+            smash-skin-e2e = 8500;
             bedwars-bow-e2e = 9000;
             # Not 10000: gameServerPort - proxyPort == 10000, so an offset of
             # 10000 aliases this gate's player port onto the base game-server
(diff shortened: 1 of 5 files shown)


andrewgazelka:
The residual wearer-own-avatar skin issue this PR isolates is tracked in #1056.

PR #1067, opened 2026-07-29
refactor(hyperion): ProjectileMotion as a per-instance component (Stage 2, inc 1)

andrewgazelka:
## What

First increment of the composable-projectile work (Stage 2). Behavior-preserving foundation: `ProjectileMotion` becomes a per-**instance** `#[derive(Component)]` instead of only a per-**kind** lookup.

- An `OnAdd(EntityKind)` observer (`seed_projectile_motion`) seeds the component from the kind's vanilla `SIMULATED` default. A kind with no entry gets nothing, exactly as before.
- `update_projectile_positions` reads the `&ProjectileMotion` component, with the kind lookup kept as a fallback.

This is the linchpin the rest of the composable work builds on: a game module can now override a single projectile's gravity/drag (a hook with no gravity, a heavier lob) without inventing a new entity kind, and the shared host-independent physics core (next increment) reads one component.

No behavior change: every simulated kind gets its existing motion.

## Verification (local, aarch64-darwin)

- `cargo clippy -p hyperion -p smash -p bedwars --all-targets -- -D warnings`: clean.
- Store-cached `checks.bedwars-bow-e2e` and `checks.smash-bow-e2e`: **both green** (the projectile path this refactors).
- `cargo nextest -p hyperion`: one unrelated pre-existing failure, `effects::status::tests::a_slow_broadcasts_to_the_victims_own_client`, which aborts the hyperion test binary with `ECS_INVALID_OPERATION: Component ...world_time::WorldTime is not registered`. **Proven pre-existing**: reverting all three of my files to origin/main reproduces the identical failure (it is `WorldTimeModule::module` calling `set::<WorldTime>` before registration — the bug being fixed separately). This PR touches only the projectile path, nothing near `world_time.rs`; the smash/bedwars servers (which register WorldTime) boot fine in the passing mkChecks.

**Do not merge until rebased onto the WorldTime registration fix** so the full hyperion suite runs green.

CI is known-broken on this repo.

### DIFF
--- a/crates/hyperion/src/egress/sync_entity_state.rs
+++ b/crates/hyperion/src/egress/sync_entity_state.rs
@@ -36,7 +36,7 @@ use crate::{
         event::{self, HitGroundEvent},
         handlers::is_grounded,
         metadata::{MetadataChanges, get_and_clear_metadata},
-        projectile_motion::{MotionOrder, lerp_rotation, look_angles},
+        projectile_motion::{MotionOrder, ProjectileMotion, lerp_rotation, look_angles},
     },
     spatial::get_first_collision,
     storage::Events,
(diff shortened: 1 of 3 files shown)


andrewgazelka:
Closing as superseded by #1076.

Checked rather than assumed. `34eb46c` on main is the expanded version of this change by the same author, over the same three files, and rebasing this branch onto current main leaves an empty diff:

- `simulation/mod.rs` auto-merges to byte-identical content, except one surviving line: `world.component::<projectile_motion::ProjectileMotion>();` inside `register_components`. That line is now actively wrong. Main registers `ProjectileMotion` in `ProjectileComponentsModule` at `projectile_motion.rs:255`, and `SimComponentsModule` imports it, so keeping the line would put one component in two registration modules and break the convention in `CLAUDE.md` that this refactor exists to establish.
- `egress/sync_entity_state.rs` and `simulation/projectile_motion.rs` conflict on comment text only. The code is identical on both sides. Main's wording is the better of the two: it names `seed_projectile_motion` as the seeding system, and records that the component is registered by `ProjectileComponentsModule`.

So there is nothing here that #1076 did not land, and one thing that would regress it. Not a judgement on the change, which is on main.

One thing worth flagging for whoever owns this branch: it is named `feat/per-event-run-apps` and carries only the projectile refactor, so the work the name promises may never have been written. If per-event run apps are still wanted, that is a fresh branch.

(sent by an AI agent via Claude Code)

PR #1096, opened 2026-07-30
feat(smash): implement the double jump (ENG-11440)

andrewgazelka:
Fixes ENG-11440.

## What was wrong

Smash had no double jump. Every part of the mechanic existed except the part
that jumps: fifteen kits declared `jump_power` and three declared
`jump_control`, `JumpsLeft` was registered, attached to every player by a
`With` trait and set to `1` in three places, and **nothing read any of it**.
No code spent a jump and no code applied an impulse.

Recovery from knockback is the whole game. A player launched off the map with
a jump left survives and one without it does not, so a mode that models the
counter and never spends it is a mode missing its defensive half.

## How it works

Vanilla creative flight with the flight taken out, which is how Mineplex's
`PerkDoubleJump` did it and needs no client mod. The server grants flight
permission while a player is airborne with a jump left; a vanilla client
answers a double tap of the jump key with a serverbound abilities packet; the
game turns that into an impulse instead of into flight.

* `server::Flight` and `Server::set_flight` are the new seam verb. Two states,
  `Armed` and `Disarmed`, rather than the protocol's two booleans: neither one
  leaves the player flying, so "keep flying" is not something this seam can be
  asked for.
* `mirror.rs` copies the host's flying bit onto a new `Flying` mirror.
* `module/jump.rs` is the behaviour, and only behaviour: `arm_double_jump`,
  `spend_double_jump`, `restore_double_jump`. It registers no components and
  imports the registration owners of every component it reads, per CLAUDE.md.
* `KitStats::jump_count`, so the Chicken's `[VERIFIED]` eight flaps are
  expressible. `kit::apply` and the respawn path read it instead of hardcoding
  one.
* `visuals::updraft` (`minecraft:small_gust` in a disc under the feet) and
  `sound::DOUBLE_JUMP` (`minecraft:entity.breeze.jump`).

Two deliberate departures from the citation, both marked in
`docs/smash-design.md` and at the constant:

1. `UtilAction.velocity`'s `yMax` of 1.0 is **not** implemented. Every
   uncontrolled kit declares `jump_power >= 0.9`, so with the `+0.2` lift a
   ceiling at 1.0 would clamp all twelve of them to an identical jump and the
   per-kit number would stop meaning anything -- which is the shape of bug this
   PR exists to fix.
2. The known "triple jump" is refused rather than reproduced: a press that
   arrives while the player is standing on something spends nothing.

## The gates, and each one watched red

**`events/smash/tests/jump.rs`**, eleven tests on the `MockServer` call log.
Broken deliberately with `jump_power: 0.0`:

```
---- a_mid_air_press_spends_one_jump_and_launches_the_player stdout ----
a double jump has to go up; got Vec3(0.0, 0.0, 0.0)
---- a_controlled_jump_goes_where_the_player_is_looking stdout ----
a controlled jump should carry the player where they look; got Vec3(0.0, 0.2, 0.0)
test result: FAILED. 9 passed; 2 failed
```

**`tests/contract.rs`**. The Player contract set `JumpsLeft(0)`, ticked and
asserted it was `1` again -- a check that passed with the entire feature
absent, because it exercised only the restore. That is now a Jump contract
that asserts a spend first. Broken deliberately by deleting `jumps.0 -= 1`:

```
every module runs with only what it declares, but these did not:
  Jump: assertion `left == right` failed: a mid-air press spent no jump
  left: 1
 right: 0
```

Note that the restore half stayed green throughout, which is exactly the hole.

**A real-client leg**, `Match.prove_double_jump` in `tools/smash-match.py`.

**This leg is what caught the one real bug in the change, and the unit tests
could not have.** The first version of the mirror computed the take-off as
`flight.is_flying && !tracking.last_tick_flying`. That edge can never fire:
hyperion decodes packets in `OnUpdate` and copies `is_flying` into
`last_tick_flying` in `PreStore`, both strictly after the smash mirror runs in
`OnLoad`, so there is no moment in the tick when the mirror can see the two
disagree. Eleven unit tests and the contract all passed, because they drive the
mirrored component directly and so never exercise the mirror at all. The double
jump was doing nothing whatsoever on a real client:

```
1.91s [P1 ] <- abilities flags=0x04 (may fly: yes)
1.93s [P1 ] -> double jump (abilities, flying)
... no SetEntityMotion
```

Green after the fix, in the leg's real position at the end of the whole
fifteen-kit sweep:

```
277.45s [P1 ] -> walking to (6.0, 65.0, -11.0)  back on the ground before the jump
278.06s [P1 ] -> walking to (6.0, 72.0, -11.0)  into the air for a double jump
279.24s PROVED a mid-air jump launched a real client: P1 sent
        ServerboundPlayerAbilities(flying) 7 blocks up and the server answered
        with SetEntityMotion (0.000, 1.000, 0.000)
```

It costs the gate about three seconds.

The leg asserts both halves in order -- the server offering `CAN_FLY` while the
player is airborne, and the launch that answers the press -- because a gate
that only watched for the launch could not tell "the jump does nothing" from
"the client was never allowed to ask".

## What was run

* `cargo test -p smash` -- 303 passed across 27 binaries.
* `cargo clippy --workspace --all-targets --all-features` -- clean.
* `nix run .#smash-e2e` -- boots the game and joins real clients. Run in both
  profiles. The dev-profile run also covers the ENG-11000 class, since a
  use-before-register would abort the boot; it cannot complete a full sweep for
  a reason that has nothing to do with this branch (ENG-11458, filed: a flecs
  generation passing 64 trips `minecraft_id`'s `debug_assert`), so the
  in-position verification above is from `HYPERION_PROFILE=release`, which is
  what the gate itself uses.
* `nix build .#checks.aarch64-darwin.smash-dev-boot-e2e` -- passed, though on a
  tree from before the mirror fix.

The release run does not come back all green, so here is the differential
against `origin/main` (d8df46f), same command, same machine:

| | this branch | origin/main |
|---|---|---|
| proved | **14 of 16** | 13 of 15 |
| `every declared ability did what it declared` | NOT PROVED | NOT PROVED |
| `match ended, back in the hub` | NOT PROVED | NOT PROVED |
| `a mid-air jump launched a real client` | **proved** | claim does not exist |

Both failures are identical on both sides. The first is `Cow / Mooshroom
Madness` and `Guardian / Target Laser declares buffs_melee and did not do it`,
word for word on both. The second is the script stalling on its `settled`
predicate -- `self.falls == len(self.deaths)`, a fall it initiated that never
produced a death -- and it is worth knowing that a match here can only end by
elimination: `LobbyConfig::default().match_timeout_seconds` is 1200 s against
the client's 300 s budget, so the timeout path is unreachable in this gate.

I ran the branch twice before taking that baseline, because a single red run is
not a finding. Both branch runs stalled at the same place (363.1 s and 365.5 s)
and both proved the double jump identically. Nothing here is a regression, and
the one column that changed changed upwards.

## Two defects found in review, fixed here

Both at the death/respawn boundary, both one root cause: a dead player is
filtered out of `arm_double_jump` and `spend_double_jump` by
`.without(RespawnAt)`, so they arrive at the respawn carrying whatever was true
when they died.

**A player who respawns still airborne had no double jump.** `MayFly` is a
write-cache of what the client was last told. Dying airborne freezes it at
`true`, while the client resets its own abilities out of band — vanilla's
`GameType.updatePlayerAbilities` sets `mayfly = false` on the gamemode change
that leaving spectator causes, and does not tell the server. `arm_double_jump`
then compared equal and sent nothing, so the player held a `JumpsLeft` their
client would refuse to spend. Not a narrow window: `respawn` reaches the host
through a teleport the client has to confirm, so `is_grounded` reports the
pre-death airborne position for several ticks. Dying in mid-air is the normal
way to die in this mode, and a fresh respawn is about to be knocked off again —
the mechanic failing in exactly the situation it exists for.

**A press made on the tick of death was replayed at the respawn.** Same cause
from the other side: the host's flying bit is still set, the adapter's
`SetFlight` is the only thing that ever clears it, and `mirror_player_state` has
no such filter, so it kept copying `is_flying = true` for the whole death
window. The first tick the filter lifted, it spent — a launch, a gust and a
sound at the spawn point from a keypress made before the player died.

The fix is three lines in `lives.rs`'s `respawn`: `MayFly(false)`,
`Flying(false)`, and `set_flight(Flight::Disarmed)`. **On the respawn transition
and deliberately not the death one** — `Op::Spectating` defers gamemode
publication to hyperion's roster, so a `Disarmed` pushed next to
`set_spectating(true)` races it and can tell a spectator it may not fly. At
respawn the two agree, because the default gamemode's abilities are
`mayfly = false` as well, so whichever packet the client applies last it ends up
agreeing with the seam.

Watched each fail on its own before fixing either. First, with nothing applied:

```
a press made before dying was replayed at the spawn point: [Vec3(0.0, 1.0, 0.0)]
```

then with only the `Flying`/`set_flight` half in:

```
a player who respawned in mid-air was never told they may jump, so the
JumpsLeft they were handed is one the client cannot spend
  left: []
 right: [Armed]
```

`a_dead_player_is_not_armed` could not see either: it asserts a spectator is
left alone and stops. Neither could the real-client leg — `prove_double_jump`
runs in the hub before anybody has died.

## Why the level-based mirror is safe

Stronger than "we clear the bit in the same tick", which is only true of the
tick. The durable guard is the `allow: false` written *alongside* it, because
`handlers.rs:440` computes `is_flying = packet.is_flying() && flight.allow`.
Once `allow` is false the client physically cannot re-raise the bit however many
packets it sends, so a level read as a press cannot repeat.

## What arming costs

Counted off the `MockServer` log; double for packets on the wire, given
ENG-11460:

| | seam calls | packets |
|---|---|---|
| an **ordinary vanilla jump**, nobody presses anything | 2 | **4** |
| a Chicken's eight flaps and the landing | 9 | **18** |

The first row is the one to know: arming is driven by leaving the ground, not by
pressing anything, so every jump every player makes now costs four abilities
packets whether or not a double jump is involved. That is the whole of why
ENG-11460 stops being harmless, and the numbers are now in that ticket.

This does not make ENG-11458 worse — the double jump mints no entities, since
`visuals::updraft` broadcasts rather than using the entity-minting emitter.

## This PR touches `CLAUDE.md`

Deliberately, and flagging it because a repo-wide instruction file is not an
ordinary file. The ordering hazard above is not discoverable from the code you
are writing when you walk into it -- you write a rising edge in a mirror, every
test goes green, and nothing tells you the expression is a constant `false`. So
it is recorded in three places, in descending order of how likely you are to be
looking at it: the module documentation of `mirror.rs` with the phase table, the
declaration of `Flying`, and one section in `CLAUDE.md` beside the existing
flecs registration convention, which is the same category of rule.

Disarm auto-merge if you would rather that section came separately.

## Filed rather than fixed

* **ENG-11458** -- the dev profile cannot survive a full smash match, as above.
* **ENG-11460** -- every `set(Flight)` puts two identical
  `ClientboundPlayerAbilities` on the wire, because the two `flecs::OnSet`
  observers at `simulation/mod.rs:608-641` each name both `Flight` and
  `FlyingSpeed` and either write fires both. Harmless until now because almost
  nothing set `Flight`; the numbers above are why it stops being.
* **ENG-11470** -- `match ended, back in the hub` has stopped proving anything,
  found while establishing that this branch did not break it.
* **ENG-11478** -- `ServerHandle` and `PlayerId` are registered only in
  `SmashModule`, a behaviour module nothing can import without a cycle, so the
  flecs import DAG does not cover them and `contract.rs::core` hand-registers
  them instead. Pre-existing (`lives.rs` has the identical dependency) and not
  fixed here, but worth the sentence it was filed with: the guard we think we
  have for that class has a manual patch under it.

## Which tree the real-client proof came from

The `SetEntityMotion (0.000, 1.000, 0.000)` above was measured on this branch
rebased onto **6b201a9**, with the respawn fix in. Main then moved again to
**94933ff** (#1098, the fleet deployment) while that run was going, so the
branch is now rebased one commit further than the tree the run compiled.

Stating it rather than glossing it, because "re-verify on the tree that will
merge" is the whole point. The proof carries in this case and the reason is
checkable: **#1098 touches no Rust at all** — `flake.nix` plus `nix/fleet/*`,
nothing under `events/`, `crates/` or `tools/` — and it modifies no line of the
e2e harness (`smash-e2e`, `runners`, `e2eOffsets`, `HYPERION_E2E`,
`gameBinaries`, `e2e.mkCheck` are all untouched). The game binary under test and
the command that drove it are identical. That is a different situation from the
6b201a9 rebase, which went through the adapter and did need a re-run.

Say the word if you would rather have the literal run on 94933ff anyway; it is
about ten minutes.

## Not verified

`smash-dev-boot-e2e` passed as a nix check, but on the tree from before the
mirror fix -- the run took long enough that the source had moved under it. The
dev-profile boot on the final tree is covered by `nix run .#smash-e2e`, which
compiles and boots the same `debug_assertions`-on binary and had real clients
join it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,3 +144,37 @@ impl Module for CombatModule {
   `cargo test` that imports your module before trusting it. Reverting a
   registration should reproduce the `ECS_INVALID_OPERATION` abort; that is the
   guard proving it works.
+
+## Mirroring host state: mirror the level, never compute an edge
+
+**The rule.** A system that copies host state onto a game component copies the
+*value*. It does not compute "this just became true". If a consumer needs an
+edge, either keep the memory somewhere that is written after the host state is,
+or mirror the level and make the consumer idempotent.
+
+**Why this exists (ENG-11440).** `events/smash/src/mirror.rs` runs in `OnLoad`.
+hyperion decodes the tick's packets in `OnUpdate` (`ingress/decode.rs`) and
+copies `is_flying` into `MovementTracking::last_tick_flying` in `PreStore`
+(`egress/sync_entity_state.rs`) -- both strictly after. So
+`bit && !hyperions_previous_bit`, evaluated in the mirror, is **always false**:
+the packet lands after the mirror has run, and by its next run hyperion's own
+"previous" has caught up. There is no point in the tick where the mirror can
+see the two disagree.
+
+**Why the test suite cannot catch it.** Every test under `events/smash/tests/`
+drives the mirrored component directly, because the mirror is host-side and a
+mock world has no host. A mirror that can only ever write `false` is therefore
+invisible to all of them: the double jump shipped eleven passing unit tests and
+a green contract while doing nothing at all on a real client. Only
+`Match.prove_double_jump` in `tools/smash-match.py` -- a real client
+double-tapping jump and getting no impulse -- showed it.
+
+**What to do.** Mirror the level, and put the idempotence in the consumer.
+`Flying` is the host's flying bit copied verbatim; `smash::Jump` answers a press
+by clearing that bit through the seam in the same tick, and refuses a press from
+a player standing on the ground. One double tap is one jump because of those two
+properties, not because the mirror was clever.
+
+**When you add a mirror of a host bit, gate it with a real client.** A Rust test
+that sets the component proves the consumer, not the mirror. The two are
+different code and only one of them has this hazard.
--- a/docs/smash-design.md
+++ b/docs/smash-design.md
@@ -245,9 +245,22 @@ Wolf and Spider). Power and height limit are per-kit sheet values.
 get another — is an artefact of the loose ground check, and the relaunch's
 changelog lists fixes in exactly that area.
 
-`events/smash/src/module/player.rs` carries `JumpsLeft` and re-arms it on
-ground contact. The launch impulse itself belongs to the host's physics, so it
-crosses the seam.
+Built, and built the same way. `events/smash/src/module/jump.rs` keeps
+`Flight::Armed` on any playing player who is off the ground with a jump left,
+pushed across the seam only when the answer changes; `src/mirror.rs` copies the
+host's flying bit onto `Flying`; and seeing it true is what spends a
+`JumpsLeft`, clears the flight and adds the velocity. No client mod, because
+the double tap a vanilla client already sends is the input.
+
+Two departures from the citation, both deliberate. `UtilAction.velocity`'s
+`yMax` of 1.0 is not implemented: every uncontrolled kit declares a
+`jump_power` of at least 0.9, so a ceiling there would clamp all twelve of them
+to the same jump and the per-kit number would stop meaning anything. And the
+"triple jump" is refused rather than reproduced — a press that arrives while
+the player is standing on something spends nothing.
+
+`jump_count` is on `KitStats` beside `jump_power`, because the Chicken's
+`[VERIFIED]` eight flaps are a per-kit number and not a constant.
 
 ---
 
@@ -508,11 +521,9 @@ Stated plainly, because these are the parts nobody can see from the code.
    ordinary relationship; the beacon, the descent and the pickup are not built.
 3. **No hunger drain.** `hunger_interval` is carried on every kit and
    `STARVE_DAMAGE` is defined, but nothing ticks it.
-4. **No double-jump impulse.** `JumpsLeft` is tracked and re-armed; applying the
-   velocity is a seam call that is not made yet.
-5. **Assists.** Only the last hit is tracked.
-6. **Seventeen of the twenty-one kits.**
-7. **Teams and the Dominate variant.**
+4. **Assists.** Only the last hit is tracked.
+5. **Seventeen of the twenty-one kits.**
+6. **Teams and the Dominate variant.**
 
 ---
 
--- a/events/smash/src/adapter.rs
+++ b/events/smash/src/adapter.rs
@@ -39,7 +39,7 @@ use hyperion::{
     },
     net::{Compose, ConnectionId, agnostic, protocol, protocol::Clientbound},
     simulation::{
-        PendingTeleportation, Velocity,
+        Flight as HostFlight, PendingTeleportation, Velocity,
         gamemode::{DefaultGamemode, Gamemode},
         metadata::living_entity::Health,
         skin::PlayerSkin,
@@ -53,8 +53,8 @@ use valence_nbt::{Compound, List, Value};
 use crate::{
     module::kit::{self, Playing},
     server::{
-        BarColour, BarSlot, BossBar, Channel, Experience, HotbarItem, Particles, PlayerId, Server,
-        SidebarLine, Sound, SoundCategory, Status, Text, Title,
+        BarColour, BarSlot, BossBar, Channel, Experience, Flight, HotbarItem, Particles, PlayerId,
+        Server, SidebarLine, Sound, SoundCategory, Status, Text, Title,
     },
 };
 
@@ -70,6 +70,10 @@ enum Op {
     Teleport {
         player: Entity,
         to: Vec3,
+    },
+    SetFlight {
+        player: Entity,
+        flight: Flight,
     },
     SetHealth {
         player: Entity,
@@ -170,6 +174,13 @@ impl Server for HyperionServer {
         self.push(Op::Teleport {
             player: entity_of(player),
             to,
+        });
+    }
+
+    fn set_flight(&self, player: PlayerId, flight: Flight) {
+        self.push(Op::SetFlight {
+            player: entity_of(player),
+            flight,
         });
     }
 
@@ -383,6 +394,27 @@ fn apply(world: WorldRef<'_>, compose: &Compose, op: Op, bars: &HashMap<Entity, 
                 return;
             }
             entity.set(PendingTeleportation::new(to));
+        }
+        Op::SetFlight { player, flight } => {
+            let entity = world.entity_from_id(player);
+            if !entity.is_alive() {
+                return;
+            }
+            // `set` and not `get::<&mut _>`: hyperion sends the abilities
+            // packet from an `OnSet` observer on this component, so a mutation
+            // that skipped the hook would change what the server believes and
+            // tell the client nothing.
+            //
+            // `is_flying` is false in both branches. The serverbound half of
+            // this exchange is the only thing that ever turns it on, and
+            // clearing it here is what ends a take-off the tick it starts:
+            // without it the client stays in creative flight and the double
+            // tap that starts the next jump never happens, because it is
+            // already flying.
+            entity.set(HostFlight {
+                allow: flight.is_armed(),
+                is_flying: false,
+            });
         }
         Op::SetHealth {
             player,
--- a/events/smash/src/lib.rs
+++ b/events/smash/src/lib.rs
@@ -27,10 +27,10 @@ use hyperion_clap::hyperion_command::CommandRegistry;
 
 use crate::module::{
     ability::AbilityModule, arena::ArenaModule, build_stamp::BuildStampModule,
-    damage::DamageModule, effect::EffectModule, hud::HudModule, kit::KitModule, kits::StockKits,
-    knockback::KnockbackModule, lives::LivesModule, lobby::LobbyModule, player::PlayerModule,
-    projectile::ProjectileModule, scoreboard::ScoreboardModule, selector::SelectorModule,
-    sound::SoundModule,
+    damage::DamageModule, effect::EffectModule, hud::HudModule, jump::JumpModule, kit::KitModule,
+    kits::StockKits, knockback::KnockbackModule, lives::LivesModule, lobby::LobbyModule,
+    player::PlayerModule, projectile::ProjectileModule, scoreboard::ScoreboardModule,
+    selector::SelectorModule, sound::SoundModule,
 };
 
 /// The whole game.
@@ -63,6 +63,10 @@ impl Module for SmashModule {
         world.import::<KitModule>();
         world.import::<ArenaModule>();
         world.import::<LivesModule>();
+        // After `Lives`: the double jump reads the two components that say a
+        // player is spectating rather than playing, so that it can leave them
+        // alone.
+        world.import::<JumpModule>();
         world.import::<ProjectileModule>();
         world.import::<LobbyModule>();
         world.import::<ScoreboardModule>();
--- a/events/smash/src/mirror.rs
+++ b/events/smash/src/mirror.rs
@@ -8,12 +8,12 @@
 
 use flecs_ecs::prelude::*;
 use glam::Vec3;
-use hyperion::simulation::{MovementTracking, Pitch, Yaw};
+use hyperion::simulation::{Flight, MovementTracking, Pitch, Yaw};
 use hyperion_inventory::PlayerInventory;
 
 use crate::{
     input::hotbar_slot,
-    module::player::{Facing, OnGround, Player, Position, SelectedSlot, Velocity},
+    module::player::{Facing, JumpPressed, OnGround, Player, Position, SelectedSlot, Velocity},
 };
 
 #[derive(Component)]
@@ -29,11 +29,13 @@ impl Module for MirrorModule {
                 &Yaw,
                 &Pitch,
                 &MovementTracking,
+                &Flight,
                 &PlayerInventory,
                 &mut Position,
                 &mut Facing,
                 &mut OnGround,
                 &mut Velocity,
+                &mut JumpPressed,
                 &mut SelectedSlot,
             )>("mirror_player_state")
             .kind(id::<flecs::pipeline::OnLoad>())
@@ -44,11 +46,13 @@ impl Module for MirrorModule {
                     yaw,
                     pitch,
                     tracking,
+                    flight,
                     inventory,
                     position,
                     facing,
                     ground,
                     velocity,
+                    jump_pressed,
                     selected,
                 )| {
                     position.0 = **source;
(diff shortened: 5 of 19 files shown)


andrewgazelka:
## Handoff for ENG-11450 (`dead-stats`): the `kit_stats.rs` patch, verified

This cannot go in this PR. `events/smash/tests/kit_stats.rs` is not on `main` —
it exists only on `feat/eng-11450-regen-and-hunger`, and it depends on `Regen`,
`Hunger`, `Gate` and `Side`, none of which exist here. Putting it in would mean
importing that whole change. So it is written and verified below instead, for
whoever lands second to apply.

### It is more than the helper: my tenth field breaks that file's compile

`every_kit_stats_field_is_gated_by_a_test_in_this_file` destructures `KitStats`
exhaustively, on purpose. `jump_count` is a tenth field, so merging this PR and
that branch in either order gives:

```
error[E0027]: pattern does not mention field `jump_count`
error: could not compile `smash` (test "kit_stats") due to 1 previous error
```

That guard worked exactly as designed — it refused to let a stat in without a
gate. It just means the handoff is four edits, not one.

### Verified, not proposed

I merged the two branches on a throwaway local branch, applied the patch and ran
it. **21 passed, 0 failed, 0 ignored**, up from 18 with two ignored.

The three `#[should_panic]` guards are the "watched it fail" evidence, and they
are permanent rather than a manual break I did once: each runs its own scenario
against a pair whose jump field is *equal* and requires the differ to panic with
`nothing in the game reads that field`. I also confirmed they are not vacuous —
repointing one at `jump_power` so it should not panic gives

```
test the_jump_power_gate_reds_when_jump_power_is_the_field_left_equal - should panic ... FAILED
note: test did not panic as expected at events/smash/tests/kit_stats.rs:431:4
```

### The `kit::apply` conflict

Same statement chain, resolves to all four setters:

```rust
        .set(Health::full(stats.max_health))
        .set(Regen(stats.regen))
        // A fresh bar, because choosing a kit is choosing its drain rate and
        // half an old kit's clock is nobody's number.
        .set(Hunger::full(stats.hunger_interval))
        .set(JumpsLeft(stats.jump_count));
```

### The patch

<details>
<summary><code>events/smash/tests/kit_stats.rs</code></summary>

```diff
--- a/events/smash/tests/kit_stats.rs
+++ b/events/smash/tests/kit_stats.rs
@@ -29,7 +29,7 @@
         kit::{self, KitStats},
         knockback::Knockback,
         lobby::{Lobby, Phase},
-        player::{Energy, Health, OnGround, Position},
+        player::{Energy, Flying, Health, OnGround, Position},
         vitals::Hunger,
     },
     server::{PlayerId, mock::Call},
@@ -338,25 +338,22 @@
     });
 }
 
-/// What ENG-11440 has to make happen.
+/// One mid-air jump, as the mirror reports one.
 ///
-/// There is no entry point for a double jump in the game at all today -- the
-/// press is not routed, `JumpsLeft` is re-armed and never spent, and no
-/// velocity is applied -- so there is nothing this function can call. It is the
-/// one line ENG-11440 replaces, and the two tests below are `#[ignore]` until
-/// it does.
+/// `Flying` is `src/mirror.rs`'s copy of the host's flying bit, which a vanilla
+/// client turns on by double-tapping the jump key -- the only mid-air jump
+/// input a client without a mod can produce. Setting it here is the press.
+/// `kit::apply` has already filled `JumpsLeft` from the kit's own `jump_count`,
+/// so there is nothing else to arrange.
 fn double_jump(gate: &Gate, side: &Side) {
-    // Airborne, which is the only precondition the game already models.
-    gate.view(side.player).set(OnGround(false));
+    let player = gate.view(side.player);
+    player.set(OnGround(false));
+    player.set(Flying(true));
 }
 
 /// `jump_power`: the double jump's impulse.
 ///
-/// `#[ignore]`: fails until ENG-11440 lands, because nothing applies the
-/// impulse. That failure *is* the finding -- do not delete the test to make the
-/// suite green.
 #[test]
-#[ignore = "ENG-11440: double jump is not implemented; nothing reads jump_power"]
 fn jump_power_changes_how_high_a_double_jump_goes() {
     let gate = Gate::new(|stats| stats.jump_power = base().jump_power * 2.0);
 
@@ -370,9 +367,7 @@
 
 /// `jump_control`: whether the double jump goes where you look.
 ///
-/// `#[ignore]`: see [`jump_power_changes_how_high_a_double_jump_goes`].
 #[test]
-#[ignore = "ENG-11440: double jump is not implemented; nothing reads jump_control"]
 fn jump_control_changes_which_way_a_double_jump_goes() {
     let gate = Gate::new(|stats| stats.jump_control = !base().jump_control);
 
@@ -386,10 +381,98 @@
     gate.advance(0.1);
 
     gate.differ("the direction the double jump asked for", |gate, side| {
+        observable_vec(gate.game.server.total_velocity(gate.id(side.player)))
+    });
+}
+
+/// `jump_count`: how many mid-air jumps the kit gets before it lands.
+///
+/// One for everybody except the Chicken, which has eight. Two presses rather
+/// than one, because a single press cannot tell the two kits apart -- both have
+/// a jump to spend on the first one, and the whole content of the field is
+/// whether there is a second.
+#[test]
+fn jump_count_changes_how_many_double_jumps_you_get() {
+    let gate = Gate::new(|stats| stats.jump_count = base().jump_count + 1);
+
+    gate.each(double_jump);
+    gate.advance(0.1);
+    // The mirror clears the bit once the game has answered, so a second press
+    // is a second write and not a continuation of the first.
+    gate.each(|gate, side| {
+        gate.view(side.player).set(Flying(false));
+    });
+    gate.advance(0.1);
+    gate.each(double_jump);
+    gate.advance(0.1);
+
+    gate.differ("the impulse two presses asked for", |gate, side| {
         observable_vec(gate.game.server.total_velocity(gate.id(side.player)))
     });
 }
 
+/// The jump gates red when the jump field is the one left equal.
+///
+/// Same half of the deletion check the `regen` and `hunger_interval` pairs
+/// make, for the three fields ENG-11440 implements. Each runs its own scenario
+/// verbatim against a pair whose jump field is equal and whose `melee_damage`
+/// differs instead -- a field nothing in a jump scenario can touch, because
+/// nobody swings.
+///
+/// Without these, a jump test whose difference came from somewhere else would
+/// pass with `jump_power` unread, which is the exact state ENG-11440 found the
+/// game in.
+#[test]
+#[should_panic(expected = "nothing in the game reads that field")]
+fn the_jump_power_gate_reds_when_jump_power_is_the_field_left_equal() {
+    let gate = Gate::new(|stats| stats.melee_damage = base().melee_damage * 2.0);
+
+    gate.each(double_jump);
+    gate.advance(0.1);
+
+    gate.differ("the impulse the double jump asked for", |gate, side| {
+        observable_vec(gate.game.server.total_velocity(gate.id(side.player)))
+    });
+}
+
+/// See [`the_jump_power_gate_reds_when_jump_power_is_the_field_left_equal`].
+#[test]
+#[should_panic(expected = "nothing in the game reads that field")]
+fn the_jump_control_gate_reds_when_jump_control_is_the_field_left_equal() {
+    let gate = Gate::new(|stats| stats.melee_damage = base().melee_damage * 2.0);
+
+    gate.each(|gate, side| {
+        gate.view(side.player)
+            .set(smash::module::player::Facing(Vec3::X));
+        double_jump(gate, side);
+    });
+    gate.advance(0.1);
+
+    gate.differ("the direction the double jump asked for", |gate, side| {
+        observable_vec(gate.game.server.total_velocity(gate.id(side.player)))
+    });
+}
+
+/// See [`the_jump_power_gate_reds_when_jump_power_is_the_field_left_equal`].
+#[test]
+#[should_panic(expected = "nothing in the game reads that field")]
+fn the_jump_count_gate_reds_when_jump_count_is_the_field_left_equal() {
+    let gate = Gate::new(|stats| stats.melee_damage = base().melee_damage * 2.0);
+
+    gate.each(double_jump);
+    gate.advance(0.1);
+    gate.each(|gate, side| {
+        gate.view(side.player).set(Flying(false));
+    });
+    gate.advance(0.1);
+    gate.each(double_jump);
+    gate.advance(0.1);
+
+    gate.differ("the impulse two presses asked for", |gate, side| {
+        observable_vec(gate.game.server.total_velocity(gate.id(side.player)))
+    });
+}
+
 // ---------------------------------------------------------------------------
 // The part that closes the class.
 // ---------------------------------------------------------------------------
@@ -407,7 +490,7 @@
 /// check anywhere notices that nothing reads it.
 #[test]
 fn every_kit_stats_field_is_gated_by_a_test_in_this_file() {
-    const FIELDS: [&str; 9] = [
+    const FIELDS: [&str; 10] = [
         "melee_damage",
         "armor",
         "knockback_taken",
@@ -416,6 +499,7 @@
         "max_health",
         "jump_power",
         "jump_control",
+        "jump_count",
         "energy",
     ];
 
@@ -428,6 +512,7 @@
         max_health: _,
         jump_power: _,
         jump_control: _,
+        jump_count: _,
         energy: _,
     } = KitStats::default();
 
```

</details>

Apply with `git apply` on `feat/eng-11450-regen-and-hunger`. `cargo test -p smash --test kit_stats` after it: 21 passed, 0 ignored.

andrewgazelka:
## ⚠️ The patch above is stale — do not apply it

Two things changed after I posted it.

**`dead-stats` rebased and rewrote the lines it targets.** `const FIELDS: [&str; 9] = [` is now `const FIELDS: &[&str] = &[` — it took #1095's `BarSlot::COUNT = ALL.len()` lesson and dropped the hand-written count, since the destructure already enforces it. So there is no `9` to bump and no `.into_iter()` to keep. The patch will not apply, and a half-applied one fails as a compile error in the one test whose job is to catch exactly this, which is a confusing way to find out.

**And the merge order inverted: this PR landed first.** `KitStats` on main now has ten fields (`jump_count`, `kit.rs:79`) and `kit_stats.rs`'s destructure names nine, so `feat/eng-11450-regen-and-hunger` will not compile on rebase:

```
error[E0027]: pattern does not mention field `jump_count`
error: could not compile `smash` (test "kit_stats") due to 1 previous error
```

That is their guard doing its job, on their side of the merge rather than mine.

## What to do instead: four edits, in place, after rebasing onto main

1. `const FIELDS: &[&str] = &[ ... ]` — add `"jump_count"`. No count change, which is the whole benefit of the slice.
2. The `let KitStats { .. }` destructure — add `jump_count: _,`.
3. `fn double_jump` — fill in the helper, and add `Flying` to the `player::{..}` import:
   ```rust
   fn double_jump(gate: &Gate, side: &Side) {
       let player = gate.view(side.player);
       player.set(OnGround(false));
       player.set(Flying(true));
   }
   ```
   `Flying` is `src/mirror.rs`'s copy of the host's flying bit, which a vanilla client raises by double-tapping jump — the only mid-air jump input a client without a mod can produce. `kit::apply` already fills `JumpsLeft` from the kit's own `jump_count`.
4. Drop the two `#[ignore]`s, and add a `jump_count` gate whose function name **begins with `jump_count`** (the sweep matches a field name as a prefix of a test name). It needs two presses: one cannot tell the two kits apart, because both have a jump to spend on the first and the entire content of the field is whether there is a second. Clear `Flying(false)` between them, since in production the mirror is what clears it once the game has answered.

The `#[should_panic]` "reds when the field is left equal" guards from the patch above are still worth having and still correct in substance — they only need the same `double_jump` helper — but they are optional where 1-4 are not.

Verified end to end before posting the original: 21 passed, 0 ignored, against 18 with two ignored.

PR #1105, opened 2026-08-03
fix(hot-reload): repair the index probe on ELF, and split the packaging so a rules edit moves one store path (ENG-12043, ENG-12035)

andrewgazelka:
Rebased onto `main` and rewritten. ENG-12035, ENG-12043, ENG-12053.

**What changed about this PR.** Its original subject, making `hyperion` a dylib so the
host and a dlopened module share one `flecs_ecs` component-index pool, **is already on
main**. #1107 was stacked on this branch and squash-merged as 59d6258, which carried this
branch's whole diff with it. Replaying the original commit onto main produces an empty
commit, and git drops it.

What did *not* reach main is that the recipe was broken on ELF, and the check that would
have said so could not run there. So this PR is now the ELF repair plus the packaging
built on top of it.

## The check passed on darwin and could not run on linux

Two failures, and the first is why nobody knew about the second.

**It could not start the script.** The runner invoked
`./crates/hyperion-hot-reload/index-probe.sh` through its `#!/usr/bin/env bash` shebang,
and the nix sandbox has no `/usr/bin/env`:

```
/nix/store/y6dfmd0...-hot-reload-index-probe/bin/hot-reload-index-probe: \
  /build/crates/hyperion-hot-reload/index-probe.sh: /usr/bin/env: bad interpreter: No such file or directory
```

darwin has that path, so the same check passed there. A gate that cannot start reports
identically to a gate that found nothing.

**The binary could not resolve `libflecs_ecs`.** With that fixed, the cargo build succeeds
and the probe does not start. The recipe set one rpath, `$target/debug`, which is right for
a workspace member's dylib and wrong for a dependency's: cargo leaves an unhashed copy of a
member's dylib in `target/debug` but a dependency's dylib only in `target/debug/deps`,
under the metadata hash `DT_NEEDED` names.

```
$ ldd target/debug/hot-reload-index-probe
    libhyperion.so => .../target/debug/libhyperion.so (0x00007f7f2bc00000)
    libflecs_ecs-c4b1a70a43230940.so => not found
```

So `libhyperion` resolved and `libflecs_ecs` did not, which is the one library the probe
exists to reason about.

**The probe's exit code was being discarded.** It runs through `| tee $out`, so the
pipeline reported tee's status. Only `grep -q PROBE_OK` stood between a failing probe and a
green check, which covers a probe that dies early but not one that prints `PROBE_OK` and
then fails. `set -o pipefail` restores it; both assertions stay.

## Verified on x86_64-linux

dev-compute-6, on this branch as rebased, through the gates rather than by hand:

```
$ nix build .#checks.x86_64-linux.hot-reload-index-probe \
            .#checks.x86_64-linux.hot-reload-source-split --no-link --print-out-paths -L
hyperion-hot-reload-index-probe> host indices: [1, 2, 3, 4] (max 4)
hyperion-hot-reload-index-probe> module's own type index: 5
hyperion-hot-reload-index-probe> hyperion::simulation::Position index: host 4, module 4
hyperion-hot-reload-index-probe> SHARED_POOL=true
hyperion-hot-reload-index-probe> SHARED_HYPERION_INDEX=true
hyperion-hot-reload-index-probe> PROBE_OK
/nix/store/h38j8hgnfg3i1gyplwzzg488g67zi5vm-hyperion-hot-reload-index-probe
/nix/store/98y15h6npdkw66h1dk76p8av7q9crl4y-hyperion-hot-reload-source-split
rc=0
```

Re-ran the probe on aarch64-darwin after the rpath change and it still prints `PROBE_OK`:
an extra rpath that resolves nothing on that platform is inert.

## Packaging: three derivations, split by source

The reload boundary is not a flag, it is which sources reach which derivation. A rules edit
**may** move the rules dylib and **must not** move the server binary, or every apply
restarts and drops every player while the pipeline stays green.

| derivation | source | moves on |
| --- | --- | --- |
| `hyperion-dylibs` | `crates/*` | an engine change |
| `smash-server` (`ExecStart`) | `crates/*` + `events/smash` | a component or engine change |
| `smash-rules` (`X-Reload-Triggers`) | the above + `events/smash-rules` | a rules change |

Measured by evaluating the three `drvPath`s, editing one file, and evaluating again:

```
edit events/smash-rules/src/lib.rs -> smash-rules                                (reload)
edit events/smash/src/lib.rs       -> smash-server, smash-rules                  (restart)
edit crates/hyperion/src/lib.rs    -> hyperion-dylibs, smash-server, smash-rules (restart)
```

`checks.hot-reload-source-split` asserts this on the source trees themselves, costing a
`diff` and no compile. Watched it fail rather than assumed: giving `smash-server` the real
rules source exits 1 with `smash-server's source carries the real rules crate. / Every
rules edit would move ExecStart and restart the server.`

Two things that were not obvious: cargo parses every workspace member before building any
of it, so a derivation cannot omit a member's directory (an empty `src/lib.rs` stub is the
fix); and `crates/hyperion-clap-macros` is a path dependency in no `members` list, so the
fileset takes `crates` as a directory.

## What is NOT done, and is tracked as ENG-12053

The packaging does **not** yet satisfy the shared-pool precondition it exists to serve.
Building all three on x86_64-linux:

- `hyperion-dylibs` builds (release, 43.39s) shipping `libhyperion.so` and both
  `libflecs_ecs*.so`.
- `smash-rules` builds: `libsmash_rules.so`, 213 KB.
- `smash-server` **does not link**: 8 undefined LMDB symbols from
  `libhyperion_permission-*.rlib` via heed/lmdb-sys, under `-C prefer-dynamic`.
- `ldd libsmash_rules.so` wants `libhyperion_hot_reload.so`, which `hyperion-dylibs` does
  not export, and `libflecs_ecs-2c5aa556f4fa166e.so`, **a different hash from the
  `c4fb60682bf29e00` shipped**. flecs was recompiled rather than reused, so the two halves
  would hold separate `INDEX_POOL`s.

The comment in `nix/hot-reload/packaging.nix` that claimed the shared target directory made
this a build-graph fact is annotated in place as intent rather than fact, rather than left
to read as true.

Also not done: no reload IPC, no tick loop, no title, no systemd wiring. `events/smash-rules`
is deliberately trivial (one system that logs) so the deployment half can be proven against
it before smash's real rules move in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3907,6 +3907,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "smash-rules"
+version = "0.1.0"
+dependencies = [
+ "flecs_ecs",
+ "hyperion-hot-reload",
+ "tracing",
+]
+
+[[package]]
 name = "snafu"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ members = [
     'crates/packet-channel',
     'events/bedwars',
     'events/smash',
+    'events/smash-rules',
     'tools/rust-mc-bot',
 ]
 resolver = '2'
--- a/crates/hyperion-hot-reload/index-probe.sh
+++ b/crates/hyperion-hot-reload/index-probe.sh
@@ -41,6 +41,13 @@ sysroot_lib="$(rustc --print sysroot)/lib/rustlib/$host_triple/lib"
 # If it comes back, so does the blanket impl.
 flags="--cfg tokio_unstable -C prefer-dynamic"
 flags="$flags -C link-arg=-Wl,-rpath,$sysroot_lib -C link-arg=-Wl,-rpath,$target/debug"
+# `deps/` as well as `debug/`, because the two dylibs are named differently
+# there. Cargo gives a workspace member's dylib an unhashed copy in
+# `target/debug` (`libhyperion.so`), but a dependency's dylib exists only in
+# `target/debug/deps` under its metadata hash -- and the DT_NEEDED entry names
+# the hashed file. So `libhyperion.so` resolved and `libflecs_ecs-<hash>.so`
+# did not, which is the one library the whole probe is about.
+flags="$flags -C link-arg=-Wl,-rpath,$target/debug/deps"
 if [ "$ext" = so ]; then
   flags="$flags -C link-arg=-Wl,--undefined-version"
 fi
--- a/events/smash-rules/Cargo.toml
+++ b/events/smash-rules/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+authors = ["Andrew Gazelka <andrew.gazelka@gmail.com>"]
+edition.workspace = true
+name = "smash-rules"
+publish = false
+version.workspace = true
+
+# A dylib and not a cdylib. `cdylib` would give this its own copy of every Rust
+# dependency, including `flecs_ecs`, and two `flecs_ecs` copies in one process
+# is two `INDEX_POOL`s indexing one world two different ways --
+# `checks.hot-reload-index-probe` exists to catch exactly that.
+[lib]
+crate-type = ["dylib"]
+
+[dependencies]
+flecs_ecs = { workspace = true }
+hyperion-hot-reload = { workspace = true }
+tracing = { workspace = true }
+
+[lints]
+workspace = true
--- a/events/smash-rules/src/lib.rs
+++ b/events/smash-rules/src/lib.rs
@@ -1,0 +1,80 @@
+//! smash's reloadable rules.
+//!
+//! This crate is the unit of hot reload: its store path is what
+//! `nix/modules/game-server.nix` puts in `X-Reload-Triggers`, so an edit here
+//! reaches a running server as `systemctl reload` and an edit anywhere else
+//! reaches it as a restart. That boundary is the source split and nothing else
+//! -- see `docs/hot-reload.md`, "Packaging: three derivations".
+//!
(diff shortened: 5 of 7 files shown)


andrewgazelka:
## Further verification

**The dev-profile boot gate passes.** This is the one that can see a
use-before-register, since a release build compiles the flecs assert out:

```
$ nix build --no-link .#checks.aarch64-darwin.smash-dev-boot-e2e
building '/nix/store/fvy88k33gw5n9yvhgc1jr8dfj5nwdkmf-hyperion-smash-dev-boot-e2e.drv'...
rc=0
```

A real client joins a `debug_assertions` smash server built with
`crate-type = ["rlib", "dylib"]`.

**Linux is still unverified, and now for a second reason.** Both builders in
`/etc/nix/machines` are out of action:

* `vc1-nix` (x86_64-linux) has a 100% full store and every build blocks on
  `waiting for the big garbage collector lock`. ENG-12036.
* `vm-builder` (aarch64-linux) cannot resolve DNS at all. Each lookup burns
  69 seconds and returns nothing, because its `resolv.conf` leads with
  `1.1.1.1`/`1.0.0.1`, neither reachable from that guest. ENG-12038.

```
$ ssh root@vm-builder 'time getent hosts cache.ix.dev'
(no output)
real	1m9.102s
```

nix's 5 s per-request budget expires against the first dead nameserver every
time, so the error is always `Resolving timed out after 5000 milliseconds` and
the build eventually dies naming an unrelated crate.

So `nix build .#checks.x86_64-linux.hot-reload-index-probe` still has to be run
by someone with a working Linux builder before this is trusted on a host.
`docs/hot-reload.md` records the same recipe working on dev-compute-6, so this
is re-verification rather than first proof, but it has not happened.

PR #1110, opened 2026-08-03
feat(hot-reload): split the packaging so a rules edit moves one store path (ENG-12035)

andrewgazelka:
Part of ENG-12035. Stacked on #1105 (and #1108, which makes the index probe work on
ELF); retarget to `main` as those merge.

The reload boundary is not a flag, it is which sources reach which derivation.
`nix/modules/game-server.nix` will put the rules dylib in `X-Reload-Triggers` and
the server binary in `ExecStart`, and nixpkgs reloads a unit whose `[Service]`
section is unchanged and whose reload triggers differ. So a rules edit **may** move
the rules dylib and **must not** move the server binary. If it moves both, every
apply restarts, every connected player is dropped, and every gate stays green.

## Three derivations, split by source

| derivation | source | moves on |
| --- | --- | --- |
| `hyperion-dylibs` | `crates/*` | an engine change |
| `smash-server` (`ExecStart`) | `crates/*` + `events/smash` | a component or engine change |
| `smash-rules` (`X-Reload-Triggers`) | the above + `events/smash-rules` | a rules change |

`cargoUnit` is not a route to any of them (rlib-only, and it says so in an
assertion), so these are plain cargo builds against the vendor directory the
workspace already produces, using the `-C prefer-dynamic` recipe
`checks.hot-reload-index-probe` proves.

## Two things that were not obvious

**Cargo parses every workspace member before building any of it**, so a derivation
cannot just omit a member's directory:

```
error: failed to load manifest for workspace member `events/bedwars`
Caused by: no targets specified in the manifest
```

An empty `src/lib.rs` is the smallest thing that lets cargo parse a member it will
never build, and being constant it keeps that member's real source out of the
derivation's inputs.

**`hyperion-dylibs` exports its cargo target directory** and the other two build on
top of it. That is not a speed optimisation. Two independent builds of `hyperion`
get two `-C metadata` hashes -- their source trees are different store paths -- and
the mangled symbols then do not match across the `dlopen` boundary. One target
directory makes "one copy of flecs in the process" a fact about the build graph
rather than a coincidence between two builds that happen to agree.

## Measured

Evaluate the three `drvPath`s, edit one file, evaluate again (aarch64-darwin):

```
edit events/smash-rules/src/lib.rs -> smash-rules                                (reload)
edit events/smash/src/lib.rs       -> smash-server, smash-rules                  (restart)
edit crates/hyperion/src/lib.rs    -> hyperion-dylibs, smash-server, smash-rules (restart)
```

`checks.hot-reload-source-split` asserts this on the source trees themselves, which
costs a `diff` and no compile. Watched it fail rather than assuming it works --
giving `smash-server` the real rules source makes it exit 1:

```
hyperion-hot-reload-source-split> smash-server's source carries the real rules crate.
hyperion-hot-reload-source-split> Every rules edit would move ExecStart and restart the server.
error: Cannot build '/nix/store/m5146f5...-hyperion-hot-reload-source-split.drv'.
```

## What this does NOT do

- **None of the three has been built yet.** The source split is measured; the
  compile is not. A build of all three on x86_64-linux is running on dev-compute-6
  and its result is not in this description.
- No systemd wiring, no reload IPC, no title. `events/smash-rules` is deliberately
  trivial (one system that logs) so the deployment half can be proven against it
  before smash's real rules move in.
- The `stamped` wrapper still bakes `HYPERION_BUILD_REV` into `ExecStart`, so today
  every commit still changes `[Service]`. Removing that is the next change and is
  required before any of this reloads rather than restarts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3907,6 +3907,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "smash-rules"
+version = "0.1.0"
+dependencies = [
+ "flecs_ecs",
+ "hyperion-hot-reload",
+ "tracing",
+]
+
+[[package]]
 name = "snafu"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ members = [
     'crates/packet-channel',
     'events/bedwars',
     'events/smash',
+    'events/smash-rules',
     'tools/rust-mc-bot',
 ]
 resolver = '2'
--- a/events/smash-rules/Cargo.toml
+++ b/events/smash-rules/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+authors = ["Andrew Gazelka <andrew.gazelka@gmail.com>"]
+edition.workspace = true
+name = "smash-rules"
+publish = false
+version.workspace = true
+
+# A dylib and not a cdylib. `cdylib` would give this its own copy of every Rust
+# dependency, including `flecs_ecs`, and two `flecs_ecs` copies in one process
+# is two `INDEX_POOL`s indexing one world two different ways --
+# `checks.hot-reload-index-probe` exists to catch exactly that.
+[lib]
+crate-type = ["dylib"]
+
+[dependencies]
+flecs_ecs = { workspace = true }
+hyperion-hot-reload = { workspace = true }
+tracing = { workspace = true }
+
+[lints]
+workspace = true
--- a/events/smash-rules/src/lib.rs
+++ b/events/smash-rules/src/lib.rs
@@ -1,0 +1,80 @@
+//! smash's reloadable rules.
+//!
+//! This crate is the unit of hot reload: its store path is what
+//! `nix/modules/game-server.nix` puts in `X-Reload-Triggers`, so an edit here
+//! reaches a running server as `systemctl reload` and an edit anywhere else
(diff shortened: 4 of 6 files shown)


andrewgazelka:
Update after building on x86_64-linux (dev-compute-6). Two of three derivations build;
the third does not, and the shared-pool claim in the source is not yet true. Filed as
ENG-12053 and annotated in `nix/hot-reload/packaging.nix` so the comment does not read
as fact.

- `hyperion-dylibs` **builds**: `Finished \`release\` profile [optimized] target(s) in
  43.39s`, shipping `libhyperion.so` (45 MB), `libflecs_ecs.so` and
  `libflecs_ecs-c4fb60682bf29e00.so` (33 MB each).
- `smash-rules` **builds**: `libsmash_rules.so`, 213 KB, from the seeded target directory.
- `smash-server` **does not link**: 8 undefined LMDB symbols (`mdb_dbi_open`,
  `mdb_txn_begin`, ...) referenced from `libhyperion_permission-*.rlib` via heed/lmdb-sys.
  Only the binary fails; both dylibs link.
- `ldd libsmash_rules.so` wants `libhyperion_hot_reload.so`, which `hyperion-dylibs` does
  not export at all, and `libflecs_ecs-2c5aa556f4fa166e.so`, which is a **different hash
  from the one shipped**. So flecs was recompiled rather than reused, and the two halves
  would hold separate `INDEX_POOL`s.

The source split itself is unaffected and still measured: a rules-only edit moves
`smash-rules` alone.

PR #1111, opened 2026-08-03
feat(projectile): stop arrows at terrain, and bound the voxel walk under it

andrewgazelka:
Fixes [ENG-12056](https://linear.app/indexable/issue/ENG-12056). Arrows stop at walls.

## The problem

Smash projectiles were point particles integrated with no reference to the world. They flew through walls, floors and pillars until their timer ran out, so every shot on every map was taken as if the geometry were not there — a player behind a pillar was exactly as shootable as one standing in the open.

Two bugs underneath it, both of which had to be fixed before a correct sweep was possible at all:

**`Blocks::first_collision` was unbounded in `t`.** It walked voxels until it left `y ∈ [-64, 320]`, so a near-horizontal ray in an empty world visited cells out to `i32::MAX / 2`. The only reason that was survivable is that it usually found something. `update_projectile_positions` compounded it: `Ray::new(center, velocity) * distance` where `distance = velocity.length()` asks about a segment `|v|` times too long, so hyperion's own arrows stopped as soon as *anything* was ahead of them on their heading rather than when they got there.

**The traversal started in the wrong cell for negative coordinates.** `Ray::voxel_traversal` began at `self.origin.as_ivec3()`, and the cast truncates towards zero: an origin at `x = -0.5` was read as cell `0` rather than cell `-1`. Half of any Minecraft map is at a negative coordinate.

## What this does

**One traversal, two sources of shapes.** `geometry::sweep::first_block_hit(from, to, shapes)` is the single swept-voxel implementation, generic over where the block shapes come from. `Blocks::first_collision` is a wrapper over it; so is the `Cubes` fixture smash's tests build a wall from. That is what makes a test about a `Cubes` wall evidence about the shipped traversal rather than about a second copy of it.

**`Ray::voxel_traversal_until(min, max, max_t)`** plus the `floor` fix. The bound is the segment, not a box drawn around the world.

**A read seam for terrain.** `smash::module::blocks::BlockWorld` is one method, defaulting to `OpenAir`. `Server` stays write-only — position and ground state are mirrored components because that is a per-player-per-tick hot path — but terrain cannot be mirrored (millions of blocks, a handful read per tick, an authoritative copy that already exists on the host), so it gets its own trait rather than a tenth `Server` method. The default means every existing test and the whole mock need no implementation. `terrain_seam.rs` points it at hyperion's `Blocks` on the host.

**The sweep runs before the entity search, and the search runs over the clipped segment.** That ordering is the whole of "you cannot be shot through a wall": a player behind one is not on the segment the arrow got to travel, so `nearest_target` never sees them. `spatial::get_first_collision` also gained the matching `t ≤ 1` bound on its entity side — bounding blocks without bounding entities would have biased every comparison towards entities, which is a regression arriving in the shape of a fix.

**On impact** the projectile stops on the surface, sticks for `STUCK_SECONDS`, plays `entity.arrow.hit` at the impact point, and expires. `Stuck` is a tag so `fly` excludes it outright: a projectile whose next sweep starts on the face it just hit would hit that face again every tick and replay the sound forever.

## Block shapes

**Handled:** full block collision shapes on the host path. `Blocks::first_collision` asks valence for `BlockState::collision_shapes()` per block, so a slab, a stair and a fence are hit where they actually are rather than anywhere in their cell — including the case where a segment through the top half of a slab's cell correctly misses. There is a unit test for exactly that.

**Not handled:** the `Cubes` test fixture is full cubes only. That is a fixture, not the shipped path.

## Verification

`crates/geometry` — 31 tests, all green. The traversal cases:

```
test ray::tests::traversal_starts_in_the_cell_containing_a_negative_origin ... ok
test ray::tests::a_bounded_traversal_stops_at_the_end_of_the_segment ... ok
test ray::tests::a_diagonal_traversal_steps_one_axis_at_a_time ... ok
test ray::tests::a_zero_length_ray_yields_only_the_cell_it_is_in ... ok
test sweep::tests::axis_aligned_shot_stops_at_the_near_face ... ok
test sweep::tests::every_axis_and_sign_stops_at_the_same_distance ... ok
test sweep::tests::a_diagonal_shot_meets_the_blocks_it_passes_through ... ok
test sweep::tests::a_corner_the_segment_misses_does_not_stop_it ... ok
test sweep::tests::a_fast_shot_cannot_tunnel_through_a_one_block_wall ... ok
test sweep::tests::negative_coordinates_traverse_the_same_as_positive_ones ... ok
test sweep::tests::a_segment_starting_inside_a_block_is_stopped_where_it_started ... ok
test sweep::tests::a_wall_beyond_the_end_of_the_segment_is_not_hit_yet ... ok
test sweep::tests::a_partial_shape_is_missed_where_a_full_cube_would_be_hit ... ok
test sweep::tests::the_nearest_of_several_blocks_is_the_one_reported ... ok
test sweep::tests::a_zero_length_segment_in_air_hits_nothing ... ok
test sweep::tests::open_ground_is_a_miss ... ok
```

`events/smash/tests/projectile_blocks.rs` — a whole smash world driven through its own tick, firing a 60 blocks/second arrow (three blocks a tick, so the wall sits between two consecutive endpoints and only the cells between them say it is there):

```
test an_arrow_stops_at_the_wall_instead_of_passing_through_it ... ok
test an_arrow_over_open_ground_flies_exactly_as_it_did_before ... ok
test a_player_behind_the_wall_is_not_hit_through_it ... ok
test a_player_in_front_of_the_wall_is_still_hit ... ok
test the_impact_is_audible_at_the_point_it_happened ... ok
test a_stuck_arrow_does_not_shoot_whoever_walks_into_it ... ok
test a_stuck_arrow_stops_existing ... ok
test direct_damage_is_unaffected ... ok
```

The whole of `events/smash` (about 350 tests) and the whole of `crates/hyperion` are green, and `cargo clippy -p geometry -p hyperion -p smash --all-targets --all-features -- -D warnings` is clean.

**The guards were watched failing, not merely written.** Reverting the `floor()` fix fails exactly the two tests written for it and nothing else:

```
ray::tests::traversal_starts_in_the_cell_containing_a_negative_origin
sweep::tests::negative_coordinates_traverse_the_same_as_positive_ones
test result: FAILED. 29 passed; 2 failed
```

Stubbing the sweep out to `None` fails five of the eight game-level tests, and exactly the right five — the three that stay green are open ground, the hit that should still land, and direct damage:

```
test result: FAILED. 3 passed; 5 failed
```

`a_player_behind_the_wall_is_not_hit_through_it` carries its own control in the same test: the identical shot with no wall must land, or the assertion proves nothing.

**Dev profile, not just the gates.** Booting `smash` with `debug_assertions` on found a failure every release gate is structurally blind to: `smash`'s new `ProjectileComponentsModule` collided with `hyperion::simulation::projectile_motion::ProjectileComponentsModule`, because flecs registers a module under its bare Rust type name before the body that renames it runs. In a release build flecs asserts are compiled out and the two modules silently alias. Filed as [ENG-12054](https://linear.app/indexable/issue/ENG-12054); worked around here by prefixing the type, with a comment saying it may not be shortened. After the rename the dev binary boots clean and runs for 41 s with no assert.

## What is not done

- **Per-ability reactions to a block impact** — [ENG-12055](https://linear.app/indexable/issue/ENG-12055). Every projectile does the same thing on a wall: it sticks. `Payload::on_hit` only fires on entity contact, so a Creeper Sulphur Bomb that meets a wall a metre from its target now stops dead instead of flying through and doing nothing. Neither is what anyone wants, but nothing regressed. The face normal is computed and currently discarded, which is the other half of that ticket.
- **A real-client gate.** `tools/smash-bow-check.py` drives a bow with a real client and reads the arrow off the wire; `nix run .#smash-bow-e2e` was run against this branch and its result is reported in a comment below. No *new* real-client assertion was added — the natural one ("an arrow fired straight down from a player standing on solid ground stops within a block") belongs in that script and is a follow-up.
- Fissure and Enderman's Block Toss still do not read blocks. They now have a seam to read through; `docs/smash-design.md` is updated to say so.

---

Written by Claude Opus via Claude Code. Reviewed by a human before merge, per the repo's one-approval rule.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3889,6 +3889,7 @@ dependencies = [
  "anyhow",
  "clap",
  "flecs_ecs",
+ "geometry",
  "glam",
  "hyperion",
  "hyperion-clap",
--- a/crates/geometry/src/lib.rs
+++ b/crates/geometry/src/lib.rs
@@ -1,2 +1,3 @@
 pub mod aabb;
 pub mod ray;
+pub mod sweep;
--- a/crates/geometry/src/ray.rs
+++ b/crates/geometry/src/ray.rs
@@ -63,9 +63,38 @@ impl Ray {
 
     /// Efficiently traverse through grid cells that the ray intersects using the Amanatides and Woo algorithm.
     /// Returns an iterator over the grid cells ([`IVec3`]) that the ray passes through.
+    ///
+    /// Unbounded in `t`: it runs until it leaves the box. Anything asking about
+    /// a finite length of ray -- one tick of a projectile, a player's reach --
+    /// wants [`Self::voxel_traversal_until`] instead, which is the difference
+    /// between walking four cells and walking to the edge of the coordinate
+    /// space.
     #[inline]
     pub fn voxel_traversal(&self, bounds_min: IVec3, bounds_max: IVec3) -> VoxelTraversal {
-        let current_pos = self.origin.as_ivec3();
+        self.voxel_traversal_until(bounds_min, bounds_max, f32::INFINITY)
+    }
+
+    /// The same traversal, stopping once the ray has travelled `max_t`.
+    ///
+    /// `t` is in units of [`Self::direction`], so a ray built with
+    /// [`Self::from_points`] is bounded to the segment by `max_t == 1.0`.
+    ///
+    /// The bound is a real cost, not a tidiness: an axis-aligned ray in an
+    /// empty world visits every cell out to the coordinate bound, and the only
+    /// reason that was survivable was that it usually found *something*.
+    #[inline]
+    pub fn voxel_traversal_until(
+        &self,
+        bounds_min: IVec3,
+        bounds_max: IVec3,
+        max_t: f32,
+    ) -> VoxelTraversal {
+        // `floor`, not `as_ivec3`. The cast truncates towards zero, which is
+        // the same thing for a positive coordinate and one cell out for a
+        // negative one -- so every traversal that began at a negative
+        // coordinate began in the wrong cell, and half of any Minecraft map is
+        // at a negative coordinate.
+        let current_pos = self.origin.floor().as_ivec3();
 
         // Determine stepping direction for each axis
         let step = IVec3::new(
@@ -104,6 +133,8 @@ impl Ray {
             t_delta,
             bounds_min,
             bounds_max,
+            t_entry: 0.0,
+            max_t,
         }
     }
 }
@@ -117,12 +148,21 @@ pub struct VoxelTraversal {
     t_delta: Vec3,
     bounds_min: IVec3,
     bounds_max: IVec3,
+    /// How far along the ray the current cell was entered.
+    t_entry: f32,
+    /// The last `t` worth reporting a cell for.
+    max_t: f32,
 }
 
 impl Iterator for VoxelTraversal {
     type Item = IVec3;
 
     fn next(&mut self) -> Option<Self::Item> {
+        // Past the end of the ray the caller asked about.
+        if self.t_entry > self.max_t {
+            return None;
+        }
+
         // Check if current position is within bounds
         if self.current_pos.x < self.bounds_min.x
             || self.current_pos.x > self.bounds_max.x
@@ -136,19 +176,26 @@ impl Iterator for VoxelTraversal {
 
         let current = self.current_pos;
 
-        // Determine which axis to step along (the one with minimum t_max)
+        // Determine which axis to step along (the one with minimum t_max).
+        // Exactly one axis per step, never two at once: stepping both across a
+        // shared corner would skip the two cells that meet there, which is a
+        // projectile passing through a corner it could not fit through.
         if self.t_max.x < self.t_max.y {
             if self.t_max.x < self.t_max.z {
+                self.t_entry = self.t_max.x;
                 self.current_pos.x += self.step.x;
                 self.t_max.x += self.t_delta.x;
             } else {
+                self.t_entry = self.t_max.z;
                 self.current_pos.z += self.step.z;
                 self.t_max.z += self.t_delta.z;
             }
         } else if self.t_max.y < self.t_max.z {
+            self.t_entry = self.t_max.y;
             self.current_pos.y += self.step.y;
             self.t_max.y += self.t_delta.y;
         } else {
+            self.t_entry = self.t_max.z;
             self.current_pos.z += self.step.z;
             self.t_max.z += self.t_delta.z;
         }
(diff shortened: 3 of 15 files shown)


andrewgazelka:
## `nix run .#smash-bow-e2e` on this branch

Ran the real-client bow gate against this branch. A real client joins, takes `/kit skeleton`, draws the bow and the harness reads the arrows off the wire.

```
minecraft:arrow is entity type 2
PASS /kit skeleton puts a minecraft:bow on a visible key (key 0)
PASS a full draw spawns at least one minecraft:arrow (got 10)
PASS the arrow's back-extrapolated trajectory passes through the shooter's eye, not the feet (0.00 off the eye line; a feet-origin is ~1.6, here 1.53)
PASS the arrow's wire heading is look_angles(velocity) = (-35.0, 19.1) (got (-35.2, 18.3))
PASS the wire yaw is not the shooter's raw look yaw of 35 (got -35.2)
PASS the server sends the arrow's velocity every tick (SetEntityMotion), the vanilla representation the client predicts smooth motion from (got 26)
PASS the arrow is never hard-teleported by a per-tick absolute EntityPositionSync (got 0)
PASS the velocity stream integrates to a real flight (73.5 blocks)
PASS gravity shows in the per-tick velocity (vy 0.870 -> -0.419)
PASS a short draw still fires (got 4 arrows)
PASS a full draw launches at vanilla's 3.0 blocks a tick (got 2.983)
PASS a shorter draw launches slower than a full one (0.402 < 2.983)
RESULT: ok (0 checks failed)
exit=0
```

**What this proves and what it does not.** It proves a real bow shot still works end to end on this branch: the arrow spawns, the per-tick velocity broadcast is intact, the flight integrates to 73.5 blocks and gravity is on it. That matters because `fly` now writes `flight.position` through a different path, and `crate::draw` reads it.

It does **not** prove block collision on a real client. The gate aims upward at pitch −20 and the arrow leaves the arena into open air, so nothing on that flight ever met a block — the sweep answered "clear" for all 26 ticks, correctly, and the assertion is silent about it either way.

The honest gap: the link this branch's Rust tests do not cover is `HyperionBlocks::sweep` → `Blocks::first_collision` reading real loaded chunks, because the game-half tests drive a `Cubes` fixture. That is the same shape as the mirror bug recorded in `CLAUDE.md` — eleven passing tests around a seam whose host half was never exercised. What closes it is one more assertion in `tools/smash-bow-check.py`: fire straight down (pitch 90) from a player standing on the arena floor and assert the arrow's broadcast positions stop within a block instead of continuing to fall. That is geometry-independent, cheap, and it is a follow-up rather than something in this PR.

What *is* covered here is that the dev-profile binary boots and runs with the real seam installed (41 s, no flecs assert), so `HyperionBlocks` is constructed and registered on the real path even though no arrow was fired at a wall through it.

PR #1112, opened 2026-08-03
fix(hot-reload): stop RUSTFLAGS splitting the build graph, and refuse an unresolved dylib (ENG-12053)

andrewgazelka:
Closes the first half of ENG-12053. Follows #1105.

After #1105, the rules dylib asked for a `libflecs_ecs` the engine derivation did not
ship:

```
$ ldd /nix/store/94qy8lrl...-smash-rules/lib/libsmash_rules.so
    libhyperion_hot_reload.so => not found
    libflecs_ecs-2c5aa556f4fa166e.so => not found

$ ls /nix/store/...-hyperion-dylibs/lib
    libflecs_ecs-c4fb60682bf29e00.so  libflecs_ecs.so  libhyperion.so
```

Two different `flecs_ecs` compilations is two `INDEX_POOL`s indexing one world two
different ways, which is exactly the configuration `checks.hot-reload-index-probe` exists
to reject, and nothing in the reload path detects it.

## It was not features, which is the obvious suspect

`cargo build --unit-graph -Z unstable-options` reports the **same** feature set for
`flecs_ecs` under `-p hyperion`, `-p smash` and `-p smash-rules`, down to the ordering.
Worth stating because it is the answer most people would reach for.

## It was the rpath, through `-C metadata`

Cargo folds RUSTFLAGS into each unit's `-C metadata`, and each derivation was passing its
own `-Wl,-rpath`. Building `flecs_ecs` twice, changing nothing but that flag:

```
-Wl,-rpath,/tmp/one -> libflecs_ecs-c1d9502659600761.dylib
-Wl,-rpath,/tmp/two -> libflecs_ecs-13cef1f428680a8e.dylib
```

Both coexist in one target directory, so the seeded artifacts were never being reused
either, which was the entire point of exporting them.

So RUSTFLAGS is now **one string in all three derivations**, and the rpaths are applied
after linking with `patchelf`, where they cannot reach the compiler.

## Two things that fall out

`hyperion-dylibs` now also builds `-p hyperion-hot-reload`. It never shipped that dylib,
and it has to: `AbiToken` compares the address of a static in that crate across the
boundary and refuses a module that linked its own copy.

Every artifact is now checked for unresolved libraries at build time. That is the guard
which would have caught this the day it was written instead of at `ldd` time on a dev
box, and it is the shape the original bug had: a dangling `DT_NEEDED` on that particular
library means two component-index pools, silently.

## Still open on ENG-12053

`smash-server`'s 8 undefined LMDB symbols (`mdb_dbi_open`, `mdb_txn_begin`, ...) from
`libhyperion_permission-*.rlib` via heed/lmdb-sys under `-C prefer-dynamic`. Whether this
change affects that is being built now and is not yet known; the answer will be added
here rather than assumed.

### DIFF
--- a/nix/hot-reload/packaging.nix
+++ b/nix/hot-reload/packaging.nix
@@ -104,30 +104,66 @@ let
   sysrootLib = "${rustToolchain}/lib/rustlib/${pkgs.stdenv.hostPlatform.rust.rustcTarget}/lib";
   inherit (pkgs.stdenv.hostPlatform) extensions;
 
-  # The link recipe `crates/hyperion-hot-reload/index-probe.sh` proves, with
-  # store paths where that script has `target/debug`.
+  # The link recipe `crates/hyperion-hot-reload/index-probe.sh` proves.
   #
   # `-C prefer-dynamic` is the load-bearing flag: it makes the server binary and
   # the rules dylib resolve `hyperion`, and through it the one `flecs_ecs` that
   # owns the component-index pool, to a shared image rather than each linking
   # its own static copy. `checks.hot-reload-index-probe` is what proves that
   # holds; this is only the flags.
-  rustFlags =
-    libDirs:
-    lib.concatStringsSep " " (
-      [ "--cfg tokio_unstable" "-C prefer-dynamic" ]
-      ++ map (dir: "-C link-arg=-Wl,-rpath,${dir}") libDirs
-      # `flecs_ecs`'s build script installs a version script naming four globs,
-      # and ld treats a pattern that matches nothing as an error by default.
-      ++ lib.optional pkgs.stdenv.hostPlatform.isElf "-C link-arg=-Wl,--undefined-version"
-    );
+  #
+  # THE SAME STRING IN ALL THREE DERIVATIONS, and that is load bearing rather
+  # than tidy. Cargo folds RUSTFLAGS into each unit's `-C metadata`, so a flag
+  # that differs by one character recompiles the entire graph under different
+  # mangled symbol names. Measured directly: building `flecs_ecs` twice,
+  # changing nothing but an rpath, produced
+  #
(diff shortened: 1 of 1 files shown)


andrewgazelka:
**Correction: this does not close ENG-12053.** Built on x86_64-linux (dev-compute-6) and
the hashes still differ, so the title overclaims. Three real gains and one unsolved cause.

The RUSTFLAGS difference was **a** cause, not the only one:

```
hyperion-dylibs ships: libflecs_ecs-bf77c44af1eeb9f5.so
smash-rules wants:     libflecs_ecs-7df6a3b3dd4401cb.so
```

What did improve, measured by `Compiling` lines per derivation:

```
hyperion-dylibs  353
smash-rules       12   (was ~353)
smash-server      72   (only crates outside hyperion's graph: hyper, bevy_app, reqwest, jemalloc)
```

So the seeded target directory is now genuinely reused for most of the graph. It is still
not reused for `flecs_ecs`, `flecs_ecs_sys`, `bindgen` and `hyperion-hot-reload`, which is
what leaves the hash split. My next hypothesis is that the vendored source path or some
other per-derivation input still reaches those units' fingerprints, but I have not measured
it and am not going to guess in a PR body.

`libhyperion_hot_reload.so` is now shipped, which it was not before.

**The new guard works, and I watched it fail.** That is what produced the evidence above
rather than an `ldd` on a dev box days later:

```
smash-rules>    libflecs_ecs-7df6a3b3dd4401cb.so => not found
unresolved libraries in /nix/store/cs7cwiby...-smash-rules/lib/libsmash_rules.so; see ENG-12053
```

**Consequence for merging:** `nix build .#packages.<sys>.smash-rules` now FAILS, by design,
because it refuses to emit a dylib whose `DT_NEEDED` does not resolve. Nothing in `checks`
or the fleet builds that attribute, so CI is unaffected, but a reviewer should merge this
knowing the attribute is red and that the redness is the guard doing its job rather than a
regression.

PR #1113, opened 2026-08-03
feat(hyperion): the tick rate the loop managed, and a ping that was measured (ENG-12065)

andrewgazelka:
Two numbers the tab list did not have. Both live in `crates/hyperion` rather than an event crate, because server telemetry is not a game concept — the same placement, and the same honesty rule, as `egress::server_load`.

Fixes ENG-12065
Refs ENG-12061, ENG-12062, ENG-12063, ENG-12064

## The tick rate

`egress::tab_list` owns a `TabList` singleton. The event crate writes the header, hyperion writes the footer, and the packet is broadcast **only when the rendered text changes**, with a unicast to catch a joining client up. Before this, bedwars built both halves itself and broadcast them to every player every tick; smash sent no `TabList` at all.

The footer follows `server_load`'s rule, adapted to a surface with no bar to fill — both numbers are printed, the first is what the loop did and the second is what it was paced to:

```
TPS 17.0 / 20.0
1 player online
```

A server keeping up prints them equal, so that is the only way `20.0` can appear: it has to survive `Tps::absorb`, which counts real ticks against a real clock. Before a window has closed it prints `TPS sampling` rather than guessing.

Not flecs's `frame_time_total`, which is time spent *inside* frames and excludes the sleep that holds 20 Hz — a rate derived from it reads near-infinite on an idle server.

## The ping

`roster.rs` sent `ping: 0` at join and nothing after. That was not a stale reading; **nothing had ever taken one**. The game server routed a serverbound `keep_alive` to `Route::Ignore` and never sent a clientbound one, and a vanilla client only ever *answers* a keep-alive, never starts one. Every player drew five full bars for a measurement that did not exist.

`egress::ping` now probes with a keep-alive, one outstanding at a time, and folds the answer in where it is decoded. `UPDATE_LATENCY` deltas go out when the **bar** moves, not when the millisecond does, batched into one packet for everyone who changed. Thresholds are read off the pinned 26.2 client jar (sha1 `2dc72797acbc1b63fc16a11c4ac393605f453754`, the jar `nix/minecraft-version.json` already pins), `PlayerTabOverlay.extractPingIcon`:

```java
Identifier sprite = info.getLatency() < 0 ? PING_UNKNOWN_SPRITE
    : (info.getLatency() < 150 ? PING_5_SPRITE
    : (info.getLatency() < 300 ? PING_4_SPRITE
    : (info.getLatency() < 600 ? PING_3_SPRITE
    : (info.getLatency() < 1000 ? PING_2_SPRITE : PING_1_SPRITE))));
```

Vanilla draws the icon and never the number, so the bar is the whole of what a player can see and the right granularity to resend at. The value on the wire stays the real measurement. A player with no reading yet publishes `-1`, which is the client's own "unknown" sprite — the state they are actually in — rather than the `0` that drew five bars.

## Which side answers keep-alive, and how I know

This is the one place the design could quietly lie. There is a proxy between the client and the game server; a proxy that answered keep-alives itself would leave the game server timing the *proxy*, producing a completely plausible number that measures the wrong thing, and nothing in a Rust test could tell.

**Static:** `crates/hyperion-proxy` contains no keep-alive handling at all. `rg 'keep_alive|KeepAlive' crates/hyperion-proxy/` exits 1. The player-to-server path in `player.rs` reads bytes off the socket and forwards frames without parsing an id.

**Measured**, which is what actually settles it. `tools/tab-list-check.py` answers keep-alives, watches a latency arrive, then **stops answering while continuing to send position**, then answers again:

```
PASS the joining roster carries -1 (no reading yet) and not the 0 that used to draw five full bars (got [-1])
PASS a real round trip arrives as an UPDATE_LATENCY delta once keep-alives are being answered (got [-1, -1, 58])
PASS the loopback round trip is in the client's top bucket (58 ms, must be under 150)
PASS the server probes repeatedly rather than once (2 keep-alives)
NOTE  going mute for 30 s: not answering keep-alives, still sending position
PASS with this client refusing to answer, the reading falls back to -1 -- so the thing answering keep-alives is the client and not the proxy (latencies seen while mute: [-1])
PASS keep-alives kept arriving while mute (2 more), so the fallback is a timeout and not the server having stopped probing
PASS the connection survives an unanswered keep-alive; hyperion does not disconnect over one
NOTE  answering keep-alives again
PASS the reading comes back once answers resume (got [61])
```

Nothing between the client and the game server can produce that transition. What is measured is the player.

## What the number contains, stated rather than hidden

**One whole tick of it is this server.** `recv_data` drains frames once per tick and the probe goes out at the end of one, so an answer cannot be seen before the next tick. That is the floor, not a worst case: measured on loopback, where the true round trip is under a millisecond, the readings were **58 ms and 61 ms**, against a tick of 59 ms on a loop managing 17 tps.

So a displayed ping is the player's round trip plus about a tick, biased one way: a player whose real ping is 100 ms is drawn at four bars rather than five. Never wrong by more than one step. Filed as ENG-12063 with the fix (timestamp a frame on arrival in the packet channel, not at decode); deliberately not corrected for here, since subtracting an estimated tick invents a number and could go negative, which is the client's "unknown" sentinel.

## Two things this turned up

- **ENG-12062:** the loop measures **17.0 tps against its own 20.0 target** on an idle server. That is what a tick-rate readout is for, and it was the first thing it said. Environment-specific until re-measured on Linux; the number is now on the wire, so `tab-list-check.py` prints it anywhere.
- **ENG-12061:** `SimComponentsModule` could not be imported into a bare world — `Prev` and `IgnMap` were registered only by `HyperionCore`. Fixed in the first commit, with `sim_components_module_stands_alone` as the guard.

## A bug my own tests were too clean to catch

The first run of this gate showed `TPS sampling` for twenty seconds while six unit tests passed. Entries older than `WINDOW` are dropped, so the oldest retained is always *inside* the window and the span to now is always a hair short of a whole one — instrumented, `n=83 span=4988ms`. Gating readiness on that span never opens. Except on perfectly regular ticks, where an entry lands exactly on the boundary and survives, which is precisely the 50 ms-multiple fixture every test used.

Fixed by tracking the first tick separately. `a_server_whose_ticks_jitter_still_reports_a_rate` uses deliberately irregular spacing and fails against the old code while the six original tests pass — that asymmetry is the point of the test.

## Verification

- `cargo test --workspace` green; `cargo clippy --all-targets --all-features -- -D warnings` (the args `checks.clippy` uses) green.
- **Every guard added here was watched fail.** Each was broken in place, the test re-run, and restored: `Prev` unregistered → `sim_components_module_stands_alone` red; the `Player`-implies-`Ping` trait removed → red; `Tps` unregistered → registration test red; a bucket threshold moved 150→160 → red; intervals counted as timestamps → red; `changed()` forced true → red; the window fix reverted → the jitter test red.
- `nix build .#checks.<sys>.tab-list-e2e` — the run above.
- `nix build .#checks.<sys>.smash-dev-boot-e2e .#checks.<sys>.bedwars-dev-boot-e2e` green in the **dev profile**, which is not optional here: bedwars aborted on `entity symbol inconsistent: bedwars::module::tab_list::TabListModule (provided) vs. hyperion::egress::tab_list::TabListModule (existing)`. flecs put both at one name; a release build compiles that assert out and silently skips the import instead. Hence `BedwarsTabListModule`. I did not work out flecs's exact scoping rule — a pre-existing `SpawnModule` pair in hyperion and bedwars does *not* collide — so the rename is what I verified, not a theory of why.

## Not done

- The gate asserts the tick rate is delivered, well-formed and in range. It does not prove the number *responds to load* — arranging a slow server inside the sandbox is not something this harness can do. That claim rests on the unit tests, which drive `absorb` directly.
- The one-tick offset above (ENG-12063).
- bedwars' header is now rebuilt once a second rather than every tick, because a two-decimal millisecond average moves on every one and would put the per-tick broadcast straight back. Its player-count line moved into hyperion's footer; nothing else about it changed.

---

Authored by Claude Opus 5 via Claude Code, driven and reviewed by @andrewgazelka.

### DIFF
--- a/crates/hyperion/src/common/mod.rs
+++ b/crates/hyperion/src/common/mod.rs
@@ -15,6 +15,14 @@ pub mod util;
 
 pub use command_channel::CommandChannel;
 
+/// How many ticks a second the game loop is paced to.
+///
+/// The number flecs is given in [`crate::HyperionCore`] and the ceiling
+/// [`crate::egress::tab_list`] prints the measured rate against, so the two
+/// cannot drift apart and the footer cannot quote a target the loop is not
+/// actually aiming at.
+pub const TICKS_PER_SECOND: f32 = 20.0;
+
 /// Shared data that is shared between the ECS framework and the IO thread.
 pub struct Shared {
     /// The compression level to use for the server. This is how long a packet needs to be before it is compressed.
--- a/crates/hyperion/src/egress/mod.rs
+++ b/crates/hyperion/src/egress/mod.rs
@@ -18,19 +18,23 @@ use crate::{
 pub mod boss_bar;
 mod channel;
 pub mod metadata;
+pub mod ping;
 pub mod player_join;
 pub mod server_load;
 mod stats;
 pub mod sync_chunks;
 mod sync_entity_state;
+pub mod tab_list;
 
 use boss_bar::BossBarModule;
 use channel::ChannelModule;
+use ping::PingModule;
 use player_join::PlayerJoinModule;
 use server_load::ServerLoadModule;
 use stats::StatsModule;
 use sync_chunks::SyncChunksModule;
 use sync_entity_state::EntityStateSyncModule;
+use tab_list::TabListModule;
 
 #[derive(Component)]
 pub struct EgressModule;
@@ -49,6 +53,8 @@ impl Module for EgressModule {
         world.import::<ChannelModule>();
         world.import::<BossBarModule>();
         world.import::<ServerLoadModule>();
+        world.import::<TabListModule>();
+        world.import::<PingModule>();
 
         system!("broadcast_chunk_deltas", world, &Compose, &mut Blocks,)
             .kind(id::<flecs::pipeline::OnUpdate>())
--- a/crates/hyperion/src/egress/ping.rs
+++ b/crates/hyperion/src/egress/ping.rs
@@ -1,0 +1,492 @@
+//! Per-player round trip time, and the ping bars the tab list draws from it.
+//!
+//! # There was no measurement to be stale
+//!
+//! `roster::entry_of` sent `ping: 0` and `list.rs` forwarded it once, at join,
+//! and that was the whole of it. The reason was not a forgotten update: it was
+//! that **nothing on either side of the connection ever asked**. The game
+//! server routed a serverbound `keep_alive` to `Route::Ignore` and never sent a
+//! clientbound one, so the client -- which only ever answers a keep-alive and
+//! never starts one -- had nothing to answer. `0` was not a reading taken once
+//! and left; it was the absence of a reading, drawn as a full five bars.
+//!
+//! # The proxy does not answer keep-alives, so this measures the player
+//!
+//! This is the one place the design could quietly lie. hyperion puts a proxy
+//! between the client and the game server, and a proxy that answered
+//! keep-alives itself would leave the game server timing the *proxy*, which
+//! would look like a plausible ping and be a measurement of the wrong thing.
+//!
+//! It does not. `crates/hyperion-proxy` contains no keep-alive handling at all
+//! -- the player-to-server direction in `player.rs` reads bytes off the socket
+//! and forwards the frames without parsing an id -- but that is an argument,
+//! not a measurement. `tools/tab-list-check.py` is the measurement: a real
+//! client answers keep-alives, watches a latency arrive, then **stops
+//! answering** while staying otherwise busy. The reading falls back to
+//! "unknown", and comes back when it answers again. Nothing between the client
+//! and the game server can produce that, so the thing being timed is the
+//! player.
+//!
+//! # What is in the number: one tick of it is this server
+//!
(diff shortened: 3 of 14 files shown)


andrewgazelka:
Also filed [ENG-12070](https://linear.app/indexable/issue/ENG-12070/flecs-module-name-collisions-abort-a-dev-build-and-silently-skip-the) for the class behind the `BedwarsTabListModule` rename in this PR. A flecs module name collision aborts a dev build and, because `ecs_assert` is compiled out of release, silently treats the second `import` as already done: a game missing a whole subsystem, green in every release gate.

Worth noting what I could not pin down, since it bears on any guard. `SpawnModule` exists in both `crates/hyperion/src/effects/spawn.rs` and `events/bedwars/src/module/spawn.rs`, both are imported into the same bedwars world, and it does *not* collide, before or after this change. Same shape as the pair that aborted. So flecs_ecs's actual naming rule is not what the error message implies, and the rename here is what I verified rather than a theory of why.

PR #1118, opened 2026-08-03
feat(hyperion): collide against 26.2's block shapes, not valence's 1.20.1 ones

andrewgazelka:
Closes [ENG-12066](https://linear.app/indexable/issue/ENG-12066/block-collision-shapes-are-minecraft-1201s-while-the-server-speaks-262). Third of the vanilla-arrow-parity stack, after #1111 (swept collision) and #1114 (the real-client bow check).

## What was wrong

`Blocks::first_collision` asked `valence_generated`'s `BlockState::collision_shapes()`. That table is built from valence's checked-in `extracted/blocks.json`, which is Minecraft **1.20.1**: 1003 blocks and 24135 states, against the pinned jar's 1196 and 32366. The geometry an entity stopped against was 1.20.1's while every client watching it ran 26.2.

**Not an arrow defect.** The same table decided where a *player* may stand (`has_block_collision`), whether a block may be placed into a player (`handlers.rs`), and whether a bedwars spawn is on solid ground. All six call sites are in this diff; the arrow was just the one that made anybody look.

## Where the shapes come from

Nowhere but the running game. Mojang's data generator describes a block state as an id and a property map and stops -- 1196 blocks in 26.2's `reports/blocks.json` and not one mention of collision -- because a collision shape is a `VoxelShape` constant compiled into each block class. Checked rather than assumed.

So `nix/java/VanillaShapes.java` asks the server: `Bootstrap.bootStrap()`, walk `Block.BLOCK_STATE_REGISTRY`, call `getCollisionShape(EmptyBlockGetter.INSTANCE, BlockPos.ZERO)` on every state. No `MinecraftServer` subclass, unlike `VanillaTrace`, because a collision shape is a pure function of the state: the blocks whose shape depends on a neighbour encode that in their own properties (`fence[north=true]`) rather than by reading the level.

The registry's ids are the protocol state ids `hyperion_minecraft_proto::block_state` already computes, so the table is indexable by a translated id with no name round trip.

**32366 states over 326 distinct shapes, 716 boxes**, matching `block_state::STATE_COUNT` exactly.

## What swapping it actually changed: one state, and not a reachable one

The part worth reading, and it is measured, not argued. `shapes_changed_since_1_20_1` in `crates/hyperion/src/simulation/blocks/translate.rs` walks all 24135 states a 1.20.1 world can hold and compares valence's boxes against 26.2's, through the same call the collision path makes:

```
states differing: 1 of 24135 ; distinct blocks: 1 ; solid->empty 1 ; empty->solid 0
  pitcher_crop[age=0, half=upper]
    1.20.1: [[0.3125, -0.0625, 0.3125, 0.6875, 0.1875, 0.6875]]
    26.2:   []
```

One state, and it is one no world contains: a pitcher crop only grows an upper half at age 3, so `age=0, half=upper` is unplaceable.

**So this fixes no collision anybody could have hit.** Saying otherwise would be inventing a bug to justify the change. What it does buy:

* The shapes now come from the version the client is running, and `nix flake check` fails if the committed table drifts from the jar. Before this, nothing anywhere connected the two.
* All 32366 of 26.2's states have a shape, rather than the 24135 the old table described. That is latent today -- hyperion reads its world from 1.20.1 anvil regions, so a post-1.20.1 state cannot reach the collision path yet -- and `blocks_added_since_1_20_1_have_shapes` says so in its own doc comment rather than dressing coverage up as a fix.
* The delta between the two versions is now a pinned number. A jar bump that moves another block's geometry fails `shapes_changed_since_1_20_1` and names the block.

## The tables

`crates/hyperion-minecraft-proto/src/collision_shape.rs`, generated by `nix/generate-collision-shapes.py`, synced by `nix run .#sync-minecraft-collision-shapes`, held current by `checks.<system>.minecraft-collision-shapes` -- the same four-part shape as `block_state.rs` and `particle.rs`.

Two arrays, because 32366 states share 326 box lists: `SHAPES` holds each distinct list once and `STATE_SHAPES` is one `u16` index per state, so a lookup is two loads and no search.

`[f32; 6]` and not `[f64; 6]`, which is a checked claim and not a shrug: every coordinate the game emits is a multiple of 1/32, so all 26 distinct values are exactly representable in `f32` and the table rounds nothing. The generator re-derives that on every run and refuses to emit a table where it stops being true.

## Guards, each broken in place and watched failing

| guard | broken by | what it said |
| --- | --- | --- |
| `const _: () = assert!(STATE_SHAPES.len() == block_state::STATE_COUNT ...)` | deleting one entry from `STATE_SHAPES` | `error[E0080]: evaluation panicked: the collision shape table and the block state table disagree about how many states exist` |
| `checks.*.minecraft-collision-shapes` | editing a box in the committed table | `committed collision shape table is stale; run: nix run .#sync-minecraft-collision-shapes` |
| `shapes_changed_since_1_20_1` | naming `torchflower_crop` as the expected difference | ``left: [pitcher_crop[age=0, half=upper]] / right: [torchflower_crop[age=0]]`` |
| generator's f32-exactness check | adding a `0.1` coordinate to the dump | `0.1 is not exactly representable in f32, so the table would round a block surface; CollisionBox needs to become [f64; 6]` |
| generator's version check | relabelling the dump `1.20.1` | `the shape dump is Minecraft 1.20.1 but the pin is 26.2` |
| generator's unit-cube/empty check | replacing the unit cube | `the shape table has no unit cube or no empty shape, which every version has` |
| `the_table_is_well_formed` | inverting a box's y | `shape 1 has an inside-out box: [0.0, 1.0, 0.0, 1.0, 0.0, 1.0]` |

## Also in this diff

`physicsSources` now decompiles the ten classes an arrow's collision actually goes through -- `Level`, `BlockGetter`, `ClipContext`, `EmptyBlockGetter`, `VoxelShape`, `Shapes`, `AABB`, `BlockHitResult`, `BlockBehaviour`, `Block`. None of them were in the set, which is exactly why the collision half was taken on trust while the launch half was cited line by line. The landmark checks cover `traverseBlocks` and `getCollisionShape` too, so a jar bump that moves either fails the build rather than leaving the Rust citing an empty file. That set is what the next PR in this stack reads.

`events/bedwars` loses `#![feature(exact_size_is_empty)]`: it existed only to call `.is_empty()` on valence's iterator, and a slice needs no nightly feature.

## Verification

* `cargo test --workspace`: green. Seven new tests (four in `hyperion-minecraft-proto/tests/collision_shape.rs`, three in `translate.rs`, plus the ones above).
* `cargo clippy --workspace --all-targets --all-features`: clean.
* `cargo fmt --all --check`: clean.
* Named geometry, each read out of `collision-shapes.json` before it was written down rather than from memory: stone the unit cube, `oak_slab` bottom `[0,0,0,1,0.5,1]` / top `[0,0.5,0,1,1,1]` / double a cube, soul sand `0.875`, a fence post 1.5 blocks high, air and torch and water empty.
* e2e on `aarch64-darwin`: `smash-bow-e2e`, `smash-dev-boot-e2e`, `bedwars-dev-boot-e2e`, `minecraft-collision-shapes`, `minecraft-block-states` all green.

### One red check, and it is not this change

`bedwars-bow-e2e` fails here with `an arrow that hits a block stops: it was re-sent at (7.50, 20.00, 8.50) with |v|=2.970`. I did not write that off against the 11% flake rate `nix/ci/flake-gate.nix` records for it, because it failed twice in a row. So I ran the same check on **ee0021b, this branch's rebase base with none of my commits**:

```
baseline run 1 rc=1   failed: ... it was re-sent at (2.50, 4.00, 15.50) with |v|=2.970
baseline run 2 rc=1   failed: ... it was re-sent at (11.50, 4.00, 9.50) with |v|=2.970
baseline run 3 rc=1
```

Same assertion, same `|v|=2.970`, on a tree this PR has not touched. Final tally on this machine: **base 1 pass / 3 fail, this branch 0 pass / 5 fail** -- one check failing about three times in four, against the 1-in-9 the flake record claims. Filed as [ENG-12085](https://linear.app/indexable/issue/ENG-12085).

It could not have been this change in any case: the box lists are byte-identical, in the same order, for every state a bedwars world can contain -- that is what `shapes_changed_since_1_20_1` measures -- so the arrow's collision point is bit-for-bit what it was.

---

Written by Claude Opus via Claude Code.


andrewgazelka:
Correcting the record this PR's investigation left behind, now that the pre-#1111 bisect has been run.

The note attached to this PR's clearing runs described `bedwars-bow-e2e` as a pre-existing flake. It was not a flake. This PR was right that it is not caused by the shapes change, but for a different reason than "it was already broken".

The check asserted on a *second* `AddEntity` packet, believing the server re-sends a pinned arrow. It does not: that packet is `send_subscribe_channel_packets` replaying the spawn for a client that has just subscribed, carrying whatever the arrow's velocity is at that moment. So it passed or failed on subscription timing, which is why it read 11% in the flake record and about 75% two days later on the same tree.

Pre-#1111 (`11381fe`) is 3/3 green, and the trace says why:

```
3.40s <- AddEntity id=850 at (5.50, 24.62,  3.00) motion=(0,0,3.000)
3.40s <- AddEntity id=850 at (0.50,  4.00, 40.00) motion=(0,0,0.000)
```

Muzzle at `z=3.00`, "impact" at `z=40.00`, both within 10 ms: 37 blocks downrange in one tick at 3 blocks a tick, with `z` exactly 40.00 every run while `x` and `y` vary. That is the overlong-ray teleport #1111's own commit message describes removing. **The old green was the artifact of the bug #1111 fixed**, so nothing here is a server regression.

Replaced in #1125; six forced-rebuild runs on `e637f712` pass six times. The flake-gate record is amended in #1130. Full traces on ENG-12085.

